PR #1, opened 2022-10-06
Add basic Get and Put Api signatures along with protobuf setup

G8XSU:
Add Basic api signatures
Main file to review : src/main/proto/ess.proto

This will auto-generate required grpc service stubs and request/response objects.

Apart from that it is mostly auto-generated setup for gradle project.

Things like Auth will be added once we add API for creating a "store".

Currently this api can be used with any of the approach mentioned in doc, that is "Independent State Updates", "Global State Counter", "Incremental State Updates". 
LDK will be using "Global State Counter", different applications might choose different approaches based on their needs.
Doc: https://docs.google.com/document/d/1mhTAnkXAtxPQRsn17gbEPIN5G8OW2h619f_yk8CXJTE/edit?usp=sharing

### DIFF
--- a/app/src/main/proto/vss.proto
+++ b/app/src/main/proto/vss.proto
@@ -1,0 +1,154 @@
+syntax = "proto3";
+option java_multiple_files = true;
+package org.vss;
+
+message GetObjectRequest {
+
+  // store_id is a keyspace identifier.
+  // Ref: https://en.wikipedia.org/wiki/Keyspace_(distributed_data_store)
+  // All APIs operate within a single store_id.
+  // It is up to clients to use single or multiple stores for their use-case.
+  // This can be used for client-isolation/ rate-limiting / throttling on the server-side.
+  // Authorization and billing can also be performed at the storeId level.
+  string store_id = 1;
+
+  // Key for which the value is to be fetched.
+  //
+  // Consistency Guarantee:
+  // Get(read) operations against a key are consistent reads and will reflect all previous writes,
+  // since Put/Write provides read-after-write and read-after-update consistency guarantees.
+  //
+  // Read Isolation:
+  // Get/Read operations against a key are ensured to have read-committed isolation.
+  // Ref: https://en.wikipedia.org/wiki/Isolation_(database_systems)#Read_committed
+  string key = 2;
+}
+
+message GetObjectResponse {
+
+  // Fetched value and version along with the corresponding key in the request.
+  KeyValue value = 2;
+}
+
+message PutObjectRequest {
+
+  // store_id is a keyspace identifier.
+  // Ref: https://en.wikipedia.org/wiki/Keyspace_(distributed_data_store)
+  // All APIs operate within a single store_id.
+  // It is up to clients to use single or multiple stores for their use-case.
+  // This can be used for client-isolation/ rate-limiting / throttling on the server-side.
+  // Authorization and billing can also be performed at the store_id level.
+  string store_id = 1;
+
+  // global_version is a sequence-number/version of the whole store. This can be used for versioning
+  // and ensures that multiple updates in case of multiple devices can only be done linearly, even
+  // if those updates did not directly conflict with each other based on keys/transaction_items.
+  //
+  // If present, the write will only succeed if the current server-side global_version against
+  // the store_id is same as in the request.
+  // Clients are expected to store (client-side) the global version against store_id.
+  // The request must contain their client-side value of global_version if global versioning and
+  // conflict detection is desired.
+  //
+  // For the first write of the store, global version should be '0'. If the write succeeds, clients
+  // must increment their global version (client-side) by 1.
+  // The server increments global_version (server-side) for every successful write, hence this
+  // client-side increment is required to ensure matching versions. This updated global version
+  // should be used in subsequent PutObjectRequests for the store.
+  //
+  // Requests with a conflicting version will fail with `CONFLICT_EXCEPTION` as ErrorCode.
+  optional int64 global_version = 2;
+
+  // Items to be written as a result of this PutObjectRequest.
+  //
+  // In an item, each key is supplied with its corresponding value and version.
+  // Clients can choose to encrypt the keys client-side in order to obfuscate their usage patterns.
+  // If the write is successful, the previous value corresponding to the key will be overwritten.
+  //
+  // Multiple items in transaction_items of a single PutObjectRequest are written in
+  // a database-transaction in an all-or-nothing fashion.
+  // Items in a single PutObjectRequest must have distinct keys.
+  //
+  // Clients are expected to store a version against every key.
+  // The write will succeed if the current DB version against the key is the same as in the request.
+  // When initiating a PutObjectRequest, the request should contain their client-side version for
+  // that key-value.
+  //
+  // For the first write of any key, the version should be '0'. If the write succeeds, the client
+  // must increment their corresponding key versions (client-side) by 1.
+  // The server increments key versions (server-side) for every successful write, hence this
+  // client-side increment is required to ensure matching versions. These updated key versions should
+  // be used in subsequent PutObjectRequests for the keys.
+  //
+  // Requests with a conflicting version will fail with `CONFLICT_EXCEPTION` as ErrorCode.
+  //
+  // Considerations for transactions:
+  // Transaction writes of multiple items have a performance overhead, hence it is recommended to use
+  // them only if required by the client application to ensure logic/code correctness.
+  // That is, transaction_items are not a substitute for batch-write of multiple unrelated items.
+  // When a write of multiple unrelated items is desired, it is recommended to use separate
+  // PutObjectRequests.
+  //
+  // Consistency guarantee:
+  // All PutObjectRequests are strongly consistent i.e. they provide read-after-write and
+  // read-after-update consistency guarantees.
+  repeated KeyValue transaction_items = 3;
+}
+
+message PutObjectResponse {
+}
+
+// When HttpStatusCode is not ok (200), the response `content` contains a serialized ErrorResponse
+// with the relevant ErrorCode and message
+message ErrorResponse {
+
+  // The error code uniquely identifying an error condition.
+  // It is meant to be read and understood programmatically by code that detects/handles errors by
+  // type.
+  ErrorCode error_code = 1;
+
+  // The error message containing a generic description of the error condition in English.
+  // It is intended for a human audience only and should not be parsed to extract any information
+  // programmatically. Client-side code may use it for logging only.
+  string message = 2;
+}
+
+// ErrorCodes to be used in ErrorResponse
+enum ErrorCode {
+
+  // Default protobuf Enum value. Will not be used as ErrorCode by server.
+  UNKNOWN = 0;
+
+  // CONFLICT_EXCEPTION is used when the request contains mismatched version (either key or global)
+  // in PutObjectRequest. For more info refer PutObjectRequest.
+  CONFLICT_EXCEPTION= 1;
+
+  // INVALID_REQUEST_EXCEPTION is used in the following cases:
+  //   - The request was missing a required argument.
+  //   - The specified argument was invalid, incomplete or in the wrong format.
+  //   - The request body of api cannot be deserialized into corresponding protobuf object.
+  INVALID_REQUEST_EXCEPTION = 2;
+
+  // An internal server error occurred, client is probably at no fault and can safely retry this
+  // error with exponential backoff.
+  INTERNAL_SERVER_EXCEPTION = 3;
+}
+
+message KeyValue {
+
+  // Key against which the value is stored.
+  string key = 1;
+
+  // Version field is used for key-level versioning.
+  // For first write of key, version should be '0'. If the write succeeds, clients must increment
+  // their corresponding key version (client-side) by 1.
+  // The server increments key version (server-side) for every successful write, hence this
+  // client-side increment is required to ensure matching versions. These updated key versions should
+  // be used in subsequent PutObjectRequests for the keys.
+  int64 version = 2;
+
+  // Object value in bytes which is stored (in put) and fetched (in get).
+  // Clients must encrypt this blob client-side before sending it over the wire to server in order
+  // to preserve privacy and security.
+  bytes value = 3;
+}
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,16 @@
+.protoc-version
+.idea/*
+*.ipr
+*.iws
+.settings/*
+.project
+.gradle/*
+gradle/*
+gradlew*
+build/*
+app/build/*
+app/bin/*
+app/.classpath
+app/.project
+app/.settings
+
--- a/README.md
+++ b/README.md
@@ -1,2 +1,2 @@
-# ess-server
-Encrypted Storage Service
+# vss-server
+Versioned Storage Service
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,0 +1,39 @@
+buildscript {
+    ext.gradleVersion = '7.5.1'
+    ext.protobufPlugInVersion = '0.8.12'
+    ext.protobufVersion = '3.21.7'
+    ext.jerseyVersion = '3.1.0'
+    ext.junitVersion = '5.9.0'
+}
+
+plugins {
+    id 'java'
+    id 'com.google.protobuf' version "${protobufPlugInVersion}"
+    id 'war'
+    id 'idea'
+}
+
+repositories {
+    mavenCentral()
+}
+
+idea {
+    module {
+        generatedSourceDirs.add(file("build/generated/proto/main"))
+    }
+}
+
+group 'org.vss'
+version '1.0'
+
+
+dependencies {
+    implementation "com.google.protobuf:protobuf-java:$protobufVersion"
+
+    testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
+    testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitVersion"
+}
+
+test {
+    useJUnitPlatform()
+}
--- a/settings.gradle
+++ b/settings.gradle
@@ -1,0 +1,2 @@
+rootProject.name = 'vss-server'
+include 'app'


src/main/proto/ess.proto
```diff
@@ -0,0 +1,64 @@
...
+    All API's operate within a single storeId.
+    StoreId is keyspace identifier. Ref: https://en.wikipedia.org/wiki/Keyspace_(distributed_data_store)
+    It is up to clients to use single or multiple store's for their usage.
+   */
+  string storeId = 1;
+  string key = 2;
+}
+message GetResponse {
+  bool success = 1;
+  KeyValue value = 2;
```
devrandom, line 22: is it useful to return just one value, or does it make more sense to return everything with a certain key prefix?  for VLS at least we always want to get a complete snapshot of the storage for a node when we do perform a `get`.  does LDK intend to make single item retrievals?

devrandom:
I think we should never check in binaries to our repos, including `gradle-wrapper.jar`.  I know it's convenient, but it's also a security problem.  It's easy enough to provide instructions in the README or add a build script to select the right version.  See https://stackoverflow.com/a/28473686

src/main/proto/ess.proto
G8XSU: regarding "return everything with a certain key prefix" => i thought about it but this makes for an unbounded query on server-side hence we should be avoiding it
single get does mean some additional work for client.

Regarding "list by key prefix" => this in intended to be solved by listKeys api which will be a paginated api. For LDK as well, we have use-case of reloading parts of snapshot. (This will be done after computing version difference of keys b/w client and server-side)

I had added "Batch" operation support as "future item" in doc, but that also won't solve this problem because there will still be some batch limit. Client needs to use versioning to ensure consistent state instead of relying on getting everything at once in consistent fashion.
Another reason, I was being bit conservative about batch operations is because we intend to support storing values upto ~5MB, even a 10 item get might mean 50MB response depending on client-application.

Let me know if that works.

Side note: Currently this api can be used with any of the approach mentioned in doc, that is "Independent State Updates", "Global State Counter", "Incremental State Updates". LDK will be using "Global State Counter"

devrandom: I don't see "Independent State Updates" in the doc.  However, that sounds like the approach that VLS would use.  But would be interested in hearing more.

So I think the plan with `listKeys` would work, but we will need this sooner rather than later, since otherwise there is no way to discover what is in the store.

G8XSU: Yes, LDK needs listKeys and getKeysSummary immediately as well.
Just didn't include it in this commit. Wanted close on basic get and put first. 

Heading was named slightly different. Added "Independent State Update" to that heading.

src/main/proto/ess.proto
```diff
@@ -0,0 +1,64 @@
...
+  ConflictException = 2;
+  InvalidRequestException = 3;
+  InternalServerException = 4;
+}
+
+message KeyValue {
+  string key = 1;
+
+  // Write will only succeed if current DB version against key is same as in request.
+  int64 version = 2;
```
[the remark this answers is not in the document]
G8XSU: I think from client perspective they will have to maintain a current version of object that they have in-memory or secondary store.
It could be more intuitive to start with 0 as first version as it will be same for client and they just call with their current version at-hand instead of any special handling.

Will add it to docs.

src/main/proto/ess.proto
```diff
@@ -0,0 +1,64 @@
+syntax = "proto3";
+option java_multiple_files = true;
+package org.ess;
+
+service EncryptedStorageService {
+  rpc get(GetRequest) returns (GetResponse);
+  rpc put(PutRequest) returns (PutResponse);
+  //TODO: Add getKeysSummary (Gives version numbers for all keys in a store) and list by prefix API's
+}
```
jkczyz, line 9: What is being "get" or "put"? I'd recommend following these naming conventions for both method and type names:

https://cloud.google.com/apis/design/naming_convention

src/main/proto/ess.proto
```diff
@@ -0,0 +1,64 @@
...
+  /*
+    All API's operate within a single storeId.
+    StoreId is keyspace identifier. Ref: https://en.wikipedia.org/wiki/Keyspace_(distributed_data_store)
+    It is up to clients to use single or multiple store's for their usage.
+   */
+  string storeId = 1;
+  string key = 2;
+}
+message GetResponse {
+  bool success = 1;
```
jkczyz, line 21: I don't think we need to build errors into the messages. This should be done at the gRPC level.

https://cloud.google.com/apis/design/errors
https://cloud.google.com/apis/design/errors#grpc_mapping

src/main/proto/ess.proto
```diff
@@ -0,0 +1,64 @@
...
+  ConflictException = 2;
+  InvalidRequestException = 3;
+  InternalServerException = 4;
+}
+
+message KeyValue {
+  string key = 1;
+
+  // Write will only succeed if current DB version against key is same as in request.
+  int64 version = 2;
```
jkczyz, line 58: Do all `KeyValue`s in a `PutRequest` need to have the same version or can they differ?

src/main/proto/ess.proto
```diff
@@ -0,0 +1,64 @@
...
+  optional ErrorResponse errorResponse = 3;
+}
+
+message PutRequest {
+  string storeId = 1;
+  /*
+  If present, write will only succeed if current DB globalVersion against a storeId is
+  same as in request.
+   */
+  optional int64 globalVersion = 2;
```
jkczyz, line 34: Could you flesh out all the docs in this file? At a high-level, the versioning scheme should be defined.

src/main/proto/ess.proto
G8XSU: Good point, changed it to GetObjectRequest and PutObjectRequest

src/main/proto/ess.proto
G8XSU: Yes that is one way to do it but grpc errors are quite limiting. 
Hence there is another way to do it which is also general practice.

For e.g. reason in errorInfo struct is a "string" due to which it is difficult to communicate possible error scenarios to client using enum.
And in most cases clients need to take different actions based on type of error returned.

src/main/proto/ess.proto
G8XSU: Every key will have its own version. so they can differ.

src/main/proto/ess.proto
G8XSU: Sounds good, i will write a basic explanation in docs.
Hoping to write a much more detailed api doc after impl though.

src/main/proto/ess.proto
jkczyz: `google.rpc.Status` seems quite flexible, actually. Besides an error code and message, you can provide any other data such as an enum.

https://github.com/googleapis/googleapis/blob/f88ca868801b2fb5fc59f69bde46106f0934e062/google/rpc/status.proto#L46

But I'd rather use the standard error codes and only provide additional details if there's a need to differentiate errors with the same code. Seems like `ErrorCode` already maps well to the provided codes?

https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto

Then the API can be written in terms of domain types rather than "response" types.

devrandom:
I think the request/response bodies may be complex enough that protobufs and strongly-typed bindings to them would be an improvement over JSON.  Also, more compact / performant.

devrandom:
> > I think the request/response bodies may be complex enough that protobufs and strongly-typed bindings to them would be an improvement over JSON. Also, more compact / performant.
> 
> I'm not sure why they would be? I agree let's avoid JSON, but there's not that many fields here, PUT is the only complicated one, and its really just a set of K-Version-V tuples. That could be sent in a format _way_ simpler than protobufs?

What did you have in mind?  (It's probably best to have a format that is commonly used in web API client code for various languages.)

devrandom:
It sounds like you'd end up supporting client libraries for different languages, people writing sub-standard encoders/decoders, etc.

G8XSU:
protobuf/grpc is industry standard and is simple.  
Main reasons for protobuf and grpc being: 
1. Highly performant 
2. Easier to maintain backwards compatibility (cannot emphasize enough)
3. Cross language support (Platform-agnostic)
4. Makes it easier to support http/2 and async requests (we also won't have to worry about connection/channel reuse/multiplexing)

There is no way, we should be writing custom encoding/format for this. (i thought @devrandom concern about sub-standard serde was on this.)
I also think there are reasonable grpc/protobuf third party libs available in languages other than the officially supported ones. for e.g. (hyperium/tonic)

The request/response objects are complex enough, there are two more api's yet to be added, we will be adding more fields and metadata to schema as server matures.

Also, i thought we had decided on it in Meeting Notes Oct 5th: "Will use java and grpc to build backend"

Jasonvdb:
Sorry wasn't in the meeting. Curious though what were the reasons java was chosen?

G8XSU:
@Jasonvdb 
Main reasons for java were : Native cloud support by official SDK's as well as related ecosystem for database accessors.
Rust doesn't have production ready SDK's from cloud providers yet. (Most are in beta, developer preview)
Client side will be written in rust and client-native languages, communicate with server side. 
Also, Its not a one-way door, same proto definition can be re-implemented by another service.

jkczyz:
Not sure if I have all the nuances of the conversations, but I believe gRPC exposes an HTTP interface with /v1, /v2, etc. style versioning. Resources:

- https://github.com/googleapis/googleapis/blob/18d0ce3fe39e3bde035bfc7f97e8360b46694508/google/api/http.proto#L44-L53
- https://grpc.io/blog/coreos/
- https://cloud.google.com/endpoints/docs/grpc/transcoding

devrandom:
> > There is no way, we should be writing custom encoding/format for this.
> 
> I don't really understand peoples' hesitance to do this - if we're concerned about adding fields and backwards compat and all that good stuff I may agree, but I don't think that's a concern - and for a three-tuple (where one element is fixed-length) almost anything is reasonable :)

I haven't convinced myself yet that the API will remain simple.

src/main/proto/ess.proto
```diff
@@ -0,0 +1,82 @@
...
+}
+
+message GetObjectRequest {
+  /*
+    All APIs operate within a single storeId.
+    StoreId is keyspace identifier.
+    Ref: https://en.wikipedia.org/wiki/Keyspace_(distributed_data_store)
+    It is up to clients to use single or multiple store's for their usage.
+   */
+  string storeId = 1;
```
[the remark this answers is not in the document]
G8XSU: Its for key grouping, 
If service owner wants to implement following:
* client-isolation/rate-limiting/throttling
* authorization
* billing etc. 
* Also I could delete a store which is no longer being used. 
They can do this based on this identifier.

src/main/proto/ess.proto
```diff
@@ -0,0 +1,82 @@
...
+  Unknown = 0;
+  ConflictException = 2;
+  InvalidRequestException = 3;
+  InternalServerException = 4;
+}
+
+message KeyValue {
+  string key = 1;
+
+  int64 version = 2;
```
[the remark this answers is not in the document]
G8XSU: Internally it is implemented like just another KV entry only. 
We can consider it but i preferred this minor implementation complexity on server side than on client side.

We don't want to explain to clients every time that this is how you should be using this if you want global consistent writes. From multiple discussions i have had on versioning of state, it looks like devs could miss this detail if api is not explicit about it. 
By doing so, we can give clients guidance on two popular options for using versioning. And they can simply choose based on their requirements.

(Using GlobalVersioning is optional for clients, for LDK it is required, VLS might not need it. 
For single-device access globalVersion is not required.)
Provides a simpler interface for clients using single-item writes who don't care about multi-item transactions. 
Using transactions for doing global consistent write is more of an implementation detail.

src/main/proto/ess.proto
G8XSU: Removed the status and ErrorResponse from ResponseObjects. 

Now if httpStatus ==200, then valid response content-entity contains valid GetObjectResponse serialized.
Otherwise if httpStatus != 200, then response content-entity contains ErrorResponse serialized.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,80 @@
...
+message GetObjectResponse {
+  KeyValue value = 2;
+}
+
+message PutObjectRequest {
+  string storeId = 1;
+  /*
+    If present, write will only succeed if current DB globalVersion against a storeId is
+    same as in request.
+    Client is expected to send their current global version at hand.
```
jkczyz, line 24: How does the client track the global version and when may it be different from the server?

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,80 @@
...
+
+    For first write of store, version should be '0'. If write succeeds, clients must increment
+    their global version(on device) by 1.
+
+    Requests with conflicting version will fail with [`ConflictException`] as ErrorCode.
```
jkczyz, line 29: Will this return the actual global version on failure? Or does the client need to do something to get in sync?

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,80 @@
...
+    It is up to clients to use single or multiple store's for their usage.
+   */
+  string storeId = 1;
+  string key = 2;
+}
+message GetObjectResponse {
+  KeyValue value = 2;
+}
+
+message PutObjectRequest {
```
jkczyz, line 19: Let's add some docs here explaining the versioning scheme.

app/src/main/proto/vss.proto
jkczyz, line 24: "at hand" is a little vague. We should spell out what this means.

jkczyz (commented):
In general, I think we should flesh out all the documentation and have some top-level or message-level docs explaining the versioning scheme. See comments for some details.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,80 @@
...
+    All APIs operate within a single storeId.
+    StoreId is keyspace identifier.
+    Ref: https://en.wikipedia.org/wiki/Keyspace_(distributed_data_store)
+    It is up to clients to use single or multiple store's for their usage.
+   */
+  string storeId = 1;
+  string key = 2;
+}
+message GetObjectResponse {
+  KeyValue value = 2;
```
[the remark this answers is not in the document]
G8XSU: Yes, its generally a good practice to return key as well. 
This is helpful if and when we want to have batch api in future. Makes for a consistent code. 
Also slightly easier to implement client side parallel batch calls.

app/src/main/proto/vss.proto
G8XSU: No, it does not. Returning it from here doesn't add much value and keeps api simpler (and can have some race-conditions if used without considering some edge cases.)
ConflictException indicates that their view of store is out-of-sync. And they need to re-sync their view by fetching all key-versions and global version from server. 
And for the keys having different version on server, client will need to update their view(can be done lazy).
(This is will also be provided as implemented method in vss-client)

app/src/main/proto/vss.proto
G8XSU: "Clients are expected to store version against every key locally. 
When initiating a PutRequest, request should contain current version for that key."
Let me know if this sounds better or we need to add more context.

jkczyz: Not sure I follow what is meant by "store version against every key locally". This comment is on `globalVersion`, and my understanding is there is one global version for a store.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,80 @@
...
+
+message PutObjectRequest {
+  string storeId = 1;
+  /*
+    If present, write will only succeed if current DB globalVersion against a storeId is
+    same as in request.
+    Client is expected to send their current global version at hand.
+
+    For first write of store, version should be '0'. If write succeeds, clients must increment
+    their global version(on device) by 1.
```
jkczyz, line 27: Should make it clear the version is incremented on the next put request?

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,80 @@
...
+  repeated KeyValue transactionItems = 3;
+}
+
+message PutObjectResponse {
+}
+
+/*
+  When HttpStatusCode is not ok(200), response `content` contains serialized ErrorResponse.
+ */
+message ErrorResponse{
```
jkczyz, line 53: nit: add space before `{`

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,80 @@
...
+  ErrorCode errorCode = 1;
+  string message = 2;
+}
+/*
+  ErrorCodes along with corresponding HttpStatusCodes used in response
+ */
+enum ErrorCode{
```
jkczyz, line 60: nit: add space before `{`

app/src/main/proto/vss.proto
jkczyz: Is there a method for fetching everything? A client may not have all keys if another client wrote a new key.

jkczyz (commented):
Could you go ahead and push any updated docs? Would be easier to read with new commits rather than looking at comments.

app/src/main/proto/vss.proto
G8XSU: Yes it will be in subsequent PR once we have get and put.

app/src/main/proto/vss.proto
G8XSU: Added clarification

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,84 @@
...
+message GetObjectResponse {
+  KeyValue value = 2;
+}
+
+message PutObjectRequest {
+  string storeId = 1;
+  /*
+    If present, write will only succeed if current DB globalVersion against a storeId is
+    same as in request.
+    Client is expected to global version against store. And PutRequest should contain
```
jkczyz, line 24: Looks like there's a word missing in the sentence.

jkczyz, line 24: s/PutRequest/PutObjectRequest

Same elsewhere.

jkczyz (commented):
Could you add documentation to all messages and fields? The message docs can go into detail about the usage scenario. And fields can explain anything that should be considered when setting them.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,84 @@
...
+
+    Write will only succeed if current DB version against key is same as in request.
+    Client is expected to store version against every key.
+    When initiating a PutRequest, request should contain their client-side version for that key.
+
+    For first write of key, version should be '0'. If write succeeds, clients must increment
+    their corresponding key versions(client-side) by 1. These updated key versions should be used in
+    subsequent PutRequests for the keys.
+
+    Requests with conflicting version will fail with [`ConflictException`] as ErrorCode.
```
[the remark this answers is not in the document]
G8XSU: Clients might need to sync their whole view depending on conflict and their usage. 
Conflict could be either due to key versions or global version. 
ErrorResponse should be simple and adding all this info could complicate it. Instead it is easier to re-use existing api's for the functionality.

app/src/main/proto/vss.proto
G8XSU: Done

G8XSU: Done, 👍 

Regarding above commend "The message docs can go into detail about the usage scenario. And fields can explain anything that should be considered when setting them."

I created an issue and we will add these details once we have implementation. 
https://github.com/lightningdevkit/vss-server/issues/3

jkczyz: Could we do this now? Doing it in a follow-up just adds extra review burden.

G8XSU: I have added lot more docs now for every field. But there are still many things which we could cover in here. 
I didn't want to stall this PR for docs changes and want to close on interface.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,169 @@
...
+    For first write of store, global version should be '0'. If write succeeds, clients must increment
+    their global version(client-side) by 1. This updated global version should be used in subsequent
+    PutObjectRequests for the store.
+
+    Requests with conflicting version will fail with [`ConflictException`] as ErrorCode.
+   */
+  optional int64 globalVersion = 2;
+
+  /*
+    Each key is supplied with its corresponding value and version.
```
jkczyz, line 62: The first sentence should give a short summary of what the field is.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,169 @@
...
+  When HttpStatusCode is not ok(200), response `content` contains serialized ErrorResponse with
+  relevant ErrorCode and errorCode's integer value as HttpStatusCode.
+ */
+message ErrorResponse {
+  /*
+    The error code is a string that uniquely identifies an error condition.
+    It is meant to be read and understood programmatically  by code that detects/handles errors by
+    type.
+   */
+  ErrorCode errorCode = 1;
```
jkczyz, line 108: Is this duplicative of the the HTTP status code? Or can they diverge?

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,169 @@
...
+}
+
+
+message PutObjectRequest {
+
+  string storeId = 1;
+  /*
+    If present, write will only succeed if current DB globalVersion against a storeId is
+    same as in request.
+    Client is expected to store(client-side) global version against storeId.
```
jkczyz, line 50: nit: space before parentheses

Likewise throughout.

jkczyz (commented):
I didn't comment exhaustively on it, but note my comment on using the definite article "the" where appropriate. We'll want to me a little more formal in the language used in docs than everyday informal speech.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,169 @@
...
+    And PutObjectRequest should contain their client-side value of globalVersion.
+
+    For first write of store, global version should be '0'. If write succeeds, clients must increment
+    their global version(client-side) by 1. This updated global version should be used in subsequent
+    PutObjectRequests for the store.
```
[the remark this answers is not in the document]
G8XSU: Server does increment the value(but only server-side) on every put, hence we need to increment value client-side for subsequent writes. 
Increment is required to ensure that value is same across client and server. 
I didn't understand how this conflicts with first paragraph.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,169 @@
...
+  KeyValue value = 2;
+}
+
+
+message PutObjectRequest {
+
+  string storeId = 1;
+  /*
+    If present, write will only succeed if current DB globalVersion against a storeId is
+    same as in request.
```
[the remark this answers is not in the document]
G8XSU: No, there still can be conflict at key-level version. 
Its mentioned in doc of next field.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,169 @@
...
+    Client is expected to store(client-side) global version against storeId.
+    And PutObjectRequest should contain their client-side value of globalVersion.
+
+    For first write of store, global version should be '0'. If write succeeds, clients must increment
+    their global version(client-side) by 1. This updated global version should be used in subsequent
+    PutObjectRequests for the store.
+
+    Requests with conflicting version will fail with [`ConflictException`] as ErrorCode.
```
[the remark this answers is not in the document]
G8XSU: Corrected at most places i could find. 
Let me know if i missed any.

app/src/main/proto/vss.proto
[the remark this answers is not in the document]
jkczyz: > I didn't understand how this conflicts with first paragraph.

The first paragraph states the write will only succeed if the versions are the same. But if the client increments the version, then the next write would fail with a naive reading of the first paragraph. But it actually won't, as you pointed out, since the server increments the version which wasn't clear.

It seems more intuitive if the client incremented prior to writing rather than after writing. Otherwise, the client's idea of the global version would be one ahead of the actual one. I guess it depends on how you look at it, especially given the empty case when 0 should be used.

Or better yet, I suppose we could always have the server return the global version in `PutObjectResponse` on success and dispense entirely with the complexity around having the client increment. Likewise when fetching all data. Thoughts? Could even be an opaque type, if desirable.

app/src/main/proto/vss.proto
jkczyz: Rather, my question was really why include it here if it is the same as the HTTP status code?

app/src/main/proto/vss.proto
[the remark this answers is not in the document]
G8XSU: Added more clarification in most recent commit for the same.

The reason client-side increment is after write is because write can fail. 
In increment-before-write approach, if write fails we will have to rollback local/secondary version numbers which could get messy to keep track of.
We don't want to be in situation where local/secondary store is ahead of primary storage. primary ahead of secondary is acceptable scenario. 

Client-side increment is common pattern used for optimistic locking. Regarding global version in response, there are key-versions as well that we need to increment and keep track of. 
One concern with client using version numbers directly from server response, is that it introduces additional trust requirement on server. 
If client blindly applies version numbers from response, it could introduce additional risk/attack-vector. (note that this risk is still present but limited to when we are syncing whole data. And we can add additional verifications there.)
That's why i thought it would be better to let client apply version increments independently. (whether it is used along with single or multiple servers)

app/src/main/proto/vss.proto
G8XSU: ErrorCode if not being used anywhere in model, could be difficult for client to understand how and where it is being used.
Also right now, model is agnostic of http protocol and status code.

app/src/main/proto/vss.proto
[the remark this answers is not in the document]
jkczyz: > If client blindly applies version numbers from response, it could introduce additional risk/attack-vector. (note that this risk is still present but limited to when we are syncing whole data. And we can add additional verifications there.)

What sort of verification? Checking that the version numbers returned are greater than or equal to the local versions? Or something more?

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+message PutObjectRequest {
+
+  /*
+    StoreId: All APIs operate within a single storeId.
+    StoreId is a keyspace identifier.
+    Ref: https://en.wikipedia.org/wiki/Keyspace_(distributed_data_store)
+    It is up to clients to use single or multiple stores for their use-case.
+    This can be used for client-isolation/ rate-limiting / throttling on the server-side.
+    Authorization and billing can also be performed at the storeId level.
+ */
```
jkczyz, line 49: Indentation of trailing "*/" seems inconsistent with the other fields. Could we use "//" instead? It's a bit more common.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+  string storeId = 1;
+  /*
+    If present, the write will only succeed if the current server-side globalVersion against
+    a storeId is same as in the request.
```
jkczyz, line 53: Likewise, the first sentence should say what `globalVersion` is before saying how it should be used.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+  /*
+    StoreId: All APIs operate within a single storeId.
+    StoreId is a keyspace identifier.
+    Ref: https://en.wikipedia.org/wiki/Keyspace_(distributed_data_store)
+    It is up to clients to use single or multiple stores for their use-case.
+    This can be used for client-isolation/ rate-limiting / throttling on the server-side.
+    Authorization and billing can also be performed at the storeId level.
+ */
+  string storeId = 1;
+  /*
```
jkczyz, line 51: nit: add blank line between comment and previous field for consistency

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+
+    Requests with conflicting version will fail with `ConflictException` as ErrorCode.
+   */
+  optional int64 globalVersion = 2;
+
+  /*
+    Items to be written as a result of this PutObjectRequest.
+
+    In an item, each key is supplied with its corresponding value and version.
+    Clients can choose to encrypt the keys client-side in order to obfuscate user's usage patterns.
```
jkczyz, line 71: s/user's/their

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+    If write is successful, previous value corresponding to the key will be overwritten.
```
jkczyz, line 72: s/write/the write
s/previous/the previous

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+
+    Multiple items in transactionItems of a single PutObjectRequest are written in
+    a database-transaction in all-or-nothing fashion.
```
jkczyz, line 75: s/in/in an

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+    Items in single PutObjectRequest must have distinct keys.
```
jkczyz, line 76: s/single/a single

app/src/main/proto/vss.proto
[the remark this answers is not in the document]
G8XSU: In PutResponse, I mean to check correctness one way is for client to run the exact same logic of increment. Which renders the server-responded versions pointless. 

And by returning versions in response, we don't want to encourage blindly applying those version-numbers on client-side as a practice.

In case of applying versions while syncing full data, one possibility is to verify against local version(>=) and also with version numbers from another server(==).

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+
+    Write will only succeed if current DB version against key is same as in request.
```
jkczyz, line 78: Missing a lot of articles. Rephrase:

"Writes will succeed if the current DB version against the key is the same as in the request."

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+    Client is expected to store version against every key.
```
jkczyz, line 79: s/Client is/Clients are
s/version/versions

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+    When initiating a PutObjectRequest, the request should contain their client-side version for
+    that key-value.
+
+    For first write of key, version should be '0'. If the write succeeds, clients must increment
```
jkczyz, line 83: s/first/the first
s/version/the version

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+    their corresponding key versions (client-side) by 1.
+    The server increments key version (server-side) for every successful write, hence this
```
jkczyz, line 85: s/version/versions

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+    client-side increment is required to ensure matching versions. These updated key versions should
+    be used in subsequent PutObjectRequests for the keys.
+
+    Requests with conflicting version will fail with `ConflictException` as ErrorCode.
+
+    Considerations for transactions:
+    Transaction writes of multiple items have a performance overhead, hence it is recommended to use
+    them only if required by the client application to ensure logic/code correctness.
+    That is, transactionItems are not a substitute for batch-write of multiple unrelated items.
+    When write of multiple unrelated items is desired, it is recommended to use separate
```
jkczyz, line 95: s/write/a write

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+    Consistency Guarantee:
+    All PutObjectRequests are strongly consistent i.e. they provide read-after-write and
+    read-after-update consistency guarantees.
+   */
+  repeated KeyValue transactionItems = 3;
+}
+
+message PutObjectResponse {
+}
+
```
jkczyz, line 107: nit: remove extra blank line

Likewise between other messages

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+
+/*
+  When HttpStatusCode is not ok (200), the response `content` contains serialized ErrorResponse with
```
jkczyz, line 110: s/serialized/a serialized

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+  the relevant ErrorCode and ErrorCode's integer value as HttpStatusCode.
+ */
+message ErrorResponse {
+  /*
+    The error code enum-value uniquely identifies an error condition.
```
jkczyz, line 115: nit: drop "enum-value" as it is implied

jkczyz, line 115: s/identifies/identifying

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+    It is meant to be read and understood programmatically by code that detects/handles errors by
+    type.
+   */
+  ErrorCode errorCode = 1;
+  /*
+    The error message contains a generic description of the error condition in English.
```
jkczyz, line 121: s/contains/containing

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,182 @@
...
+    It is intended for a human audience only. And should not be parsed to extract any information
```
jkczyz, line 122: s/. And/and

jkczyz (commented):
Largely looks good modulo commented on grammar and nits.

app/src/main/proto/vss.proto
G8XSU: Done

app/src/main/proto/vss.proto
G8XSU: Done

app/src/main/proto/vss.proto
G8XSU: Done consistently

app/src/main/proto/vss.proto
G8XSU: Since only single version against a key needs to be stored. 
I think current version is correct ? 
Or 
"Clients are expected to store a version against every key."

app/src/main/proto/vss.proto
G8XSU: i was following request-response separated by single blank line and other proto's by double vertical space.
replaced by single blank lines for consistency.

app/src/main/proto/vss.proto
G8XSU: i think 
"The error code uniquely identifies an error condition." is correct
same with below comment "s/contains/containing"
Is it because we want doc to explain a "following" field?

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,153 @@
...
+  //  The server increments key version (server-side) for every successful write, hence this
+  //  client-side increment is required to ensure matching versions. These updated key versions should
+  //  be used in subsequent PutObjectRequests for the keys.
+  int64 version = 2;
+
+  //  Object value in bytes which is stored (in put) and fetched (in get).
+  //  Clients must encrypt this blob client-side before sending it over the wire to server in order
+  //  to preserve privacy and security.
+  bytes value = 3;
+}
```
jkczyz, line 153: nit: add newline

app/src/main/proto/vss.proto
jkczyz: > i think "The error code uniquely identifies an error condition." is correct same with below comment "s/contains/containing"

Both are grammatically correct. It's more a preference around the phrasing used to document an item. Compare the phrasing used here vs the docs for `key` and `value` fields of `KeyValue`.

> Is it because we want doc to explain a "following" field?

I'm not sure I understand what you mean by that.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,153 @@
...
+message KeyValue {
+
+  //  Key against which the value is stored.
+  string key = 1;
+
+  //  For first write of key, version should be '0'. If the write succeeds, clients must increment
+  //  their corresponding key version (client-side) by 1.
+  //  The server increments key version (server-side) for every successful write, hence this
+  //  client-side increment is required to ensure matching versions. These updated key versions should
+  //  be used in subsequent PutObjectRequests for the keys.
```
jkczyz, line 146: First sentence should document what the version is rather than jumping into how to set it.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,153 @@
+syntax = "proto3";
+option java_multiple_files = true;
+package org.vss;
+
+message GetObjectRequest {
+
+  //  StoreId: All APIs operate within a single storeId.
```
jkczyz, line 7: Feel like this sentence should go after the next one and doesn't need to be prefixed by "StoreId:".

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,153 @@
...
+  //  Clients are expected to store (client-side) the global version against storeId.
+  //  And the request should contain their client-side value of globalVersion.
+  //
+  //  For the first write of the store, global version should be '0'. If the write succeeds, clients
+  //  must increment their global version (client-side) by 1.
+  //  The server increments globalVersion (server-side) for every successful write, hence this
+  //  client-side increment is required to ensure matching versions. This updated global version
+  //  should be used in subsequent PutObjectRequests for the store.
+  //
+  //  Requests with conflicting version will fail with `ConflictException` as ErrorCode.
```
jkczyz, line 58: s/conflicting/a conflicting

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,153 @@
...
+  //  In an item, each key is supplied with its corresponding value and version.
+  //  Clients can choose to encrypt the keys client-side in order to obfuscate their usage patterns.
+  //  If the write is successful, the previous value corresponding to the key will be overwritten.
+  //
+  //  Multiple items in transactionItems of a single PutObjectRequest are written in
+  //  a database-transaction in an all-or-nothing fashion.
+  //  Items in a single PutObjectRequest must have distinct keys.
+  //
+  //  The write will succeed if the current DB version against the key is the same as in the request.
+  //  Clients are expected to store a version against every key.
```
jkczyz, line 72: I think these two sentences would read better if the ordering was reversed.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,153 @@
...
+  //  When initiating a PutObjectRequest, the request should contain their client-side version for
+  //  that key-value.
+  //
+  //  For the first write of key, the version should be '0'. If the write succeeds, clients must
```
jkczyz, line 76: s/key/any key
s/clients/the client

app/src/main/proto/vss.proto
jkczyz, line 7: Only use one space between `//` and text throughout.

jkczyz (commented):
Looks good. Just a few minor comments.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,154 @@
...
+  // When initiating a PutObjectRequest, the request should contain their client-side version for
+  // that key-value.
+  //
+  // For the first write of any key, the version should be '0'. If the write succeeds, the client
+  // must increment their corresponding key versions (client-side) by 1.
+  // The server increments key versions (server-side) for every successful write, hence this
+  // client-side increment is required to ensure matching versions. These updated key versions should
+  // be used in subsequent PutObjectRequests for the keys.
+  //
+  // Requests with conflicting version will fail with `ConflictException` as ErrorCode.
```
jkczyz, line 82: Also, "a conflicting" here.

jkczyz (approved):
@devrandom Could you take another pass?

app/src/main/proto/vss.proto
G8XSU: That makes sense, i will remove this association from ErrorCode enum's value. 
Since http status code to errorCode could be one to many in future.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,154 @@
...
+  string storeId = 1;
+
+  // GlobalVersion is a sequence-number/version of the whole store. This can be used for versioning
+  // and ensures that multiple updates in case of multiple devices can only be done linearly, even
+  // if those updates did not directly conflict with each other based on keys/transactionItems.
+  //
+  // If present, the write will only succeed if the current server-side globalVersion against
+  // a storeId is same as in the request.
+  // Clients are expected to store (client-side) the global version against storeId.
+  // And the request should contain their client-side value of globalVersion.
```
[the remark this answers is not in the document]
G8XSU: remove it due to another change but there's nothing wrong with starting a sentence with "and"

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,154 @@
...
+  // It is up to clients to use single or multiple stores for their use-case.
+  // This can be used for client-isolation/ rate-limiting / throttling on the server-side.
+  // Authorization and billing can also be performed at the storeId level.
+  string storeId = 1;
+
+  // GlobalVersion is a sequence-number/version of the whole store. This can be used for versioning
+  // and ensures that multiple updates in case of multiple devices can only be done linearly, even
+  // if those updates did not directly conflict with each other based on keys/transactionItems.
+  //
+  // If present, the write will only succeed if the current server-side globalVersion against
```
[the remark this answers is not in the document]
G8XSU: done

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,154 @@
...
+  // All APIs operate within a single storeId.
+  // It is up to clients to use single or multiple stores for their use-case.
+  // This can be used for client-isolation/ rate-limiting / throttling on the server-side.
+  // Authorization and billing can also be performed at the storeId level.
+  string storeId = 1;
+
+  // Key for which the value is to be fetched.
+  //
+  // Consistency Guarantee:
+  // Get/Read operations against a key are consistent reads and will reflect all previous writes,
```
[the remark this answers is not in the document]
G8XSU: Done

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,154 @@
+syntax = "proto3";
+option java_multiple_files = true;
+package org.vss;
+
+message GetObjectRequest {
```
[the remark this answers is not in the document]
G8XSU: This is mainly inspired from aws s3's getObject and putObject api reference. 
Normally, we cannot be fully descriptive in api name, otherwise we will also have to consider GetKeyValueVersionRequest 

I was just treating it as blob/object store api. Let me know your thoughts on this.
I can also consider renaming value to objectValue if there's some confusion.

app/src/main/proto/vss.proto
```diff
@@ -0,0 +1,154 @@
...
+  // Consistency Guarantee:
+  // All PutObjectRequests are strongly consistent i.e. they provide read-after-write and
+  // read-after-update consistency guarantees.
+  repeated KeyValue transactionItems = 3;
+}
+
+message PutObjectResponse {
+}
+
+// When HttpStatusCode is not ok (200), the response `content` contains a serialized ErrorResponse
```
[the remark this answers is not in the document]
devrandom: I somehow missed that the current PR is not for gRPC anymore.  is there more information about the underlying protocol - I don't immediately see it in the PR?

app/src/main/proto/vss.proto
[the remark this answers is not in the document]
devrandom: OK, I don't feel very strongly about this

app/src/main/proto/vss.proto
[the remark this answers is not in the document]
devrandom: OK, sounds good

devrandom (approved):

app/src/main/proto/vss.proto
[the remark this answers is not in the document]
jkczyz: FWIW, the `Status` error codes are meant to be protocol-agnostic.

https://cloud.google.com/apis/design/errors

jkczyz (approved):

app/src/main/proto/vss.proto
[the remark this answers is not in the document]
G8XSU: Whether we use HttpStatus codes or protobuf status codes, 
the same concern remains that in future we might want to introduce error conditions which donot map directly or have one to many mapping.
Removed this association as part of another comment.

G8XSU:
Changed to underscore_separated_names for field names and kept camel case for structs. (acc. to protobuf generation guideline).
Synced with Matt offline and will be going ahead and merging this.

PR #5, opened 2023-01-13
Add initial PostgreSQL implementation for get & put operations

G8XSU:
**TestStrategy**

  How? 
  * PostgresIntegrationTest uses [testcontainers](https://github.com/testcontainers) to spin up docker containers for each test.
  * AbstractKVStoreIntegrationTest defines behavior that every impl of KVStore needs to follow. Different impl's can re-use the same abstract class to test their impl against it. 
  * We will keep on adding tests to this as we add more impl details.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,39 +1,98 @@
 buildscript {
-    ext.gradleVersion = '7.5.1'
-    ext.protobufPlugInVersion = '0.8.12'
-    ext.protobufVersion = '3.21.7'
-    ext.jerseyVersion = '3.1.0'
-    ext.junitVersion = '5.9.0'
+  ext.gradleVersion = '7.5.1'
+  ext.protobufPlugInVersion = '0.8.12'
+  ext.protobufVersion = '3.21.7'
+  ext.jerseyVersion = '3.1.0'
+  ext.junitVersion = '5.9.0'
+  ext.postgresVersion = '42.5.1'
+  ext.jooqVersion = '3.17.7'
+  ext.guiceVersion = '5.1.0'
 }
 
 plugins {
-    id 'java'
-    id 'com.google.protobuf' version "${protobufPlugInVersion}"
-    id 'war'
-    id 'idea'
+  id 'java'
+  id 'com.google.protobuf' version "${protobufPlugInVersion}"
+  id 'war'
+  id 'idea'
+  id 'nu.studer.jooq' version '8.0'
 }
 
 repositories {
-    mavenCentral()
+  mavenCentral()
 }
 
 idea {
-    module {
-        generatedSourceDirs.add(file("build/generated/proto/main"))
-    }
+  module {
+    generatedSourceDirs.add(file("build/generated/proto/main"))
+  }
 }
 
 group 'org.vss'
 version '1.0'
 
-
 dependencies {
-    implementation "com.google.protobuf:protobuf-java:$protobufVersion"
+  implementation "com.google.protobuf:protobuf-java:$protobufVersion"
 
-    testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
-    testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitVersion"
+  //jOOQ & Postgres impl deps
+  implementation "org.jooq:jooq:$jooqVersion"
+  implementation "org.jooq:jooq-meta:$jooqVersion"
+  implementation "org.jooq:jooq-codegen:$jooqVersion"
+  runtimeOnly "org.postgresql:postgresql:$postgresVersion"
+  jooqGenerator "org.postgresql:postgresql:$postgresVersion"
+
+  implementation "com.google.inject:guice:$guiceVersion"
+
+  testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
+  testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitVersion"
 }
 
 test {
-    useJUnitPlatform()
+  useJUnitPlatform()
+}
+
+protobuf {
+  protoc {
+    artifact = "com.google.protobuf:protoc:$protobufVersion"
+  }
+}
+
+jooq {
+  configurations {
+    main {
+      generateSchemaSourceOnCompilation = true
+
+      generationTool {
+        jdbc {
+          driver = 'org.postgresql.Driver'
+          url = 'jdbc:postgresql://localhost:5432/postgres'
+          user = 'postgres'
+          password = ''
+          properties {
+            property {
+              key = 'ssl'
+              value = 'false'
+            }
+          }
+        }
+        generator {
+          name = 'org.jooq.codegen.DefaultGenerator'
+          database {
+            name = 'org.jooq.meta.postgres.PostgresDatabase'
+            inputSchema = 'public'
+          }
+          generate {
+            deprecated = false
+            records = true
+            immutablePojos = true
+            fluentSetters = true
+          }
+          target {
+            packageName = 'org.vss.postgres'
+            directory = 'build/generated-src/jooq/main'
+          }
+          strategy.name = 'org.jooq.codegen.DefaultGeneratorStrategy'
+        }
+      }
+    }
+  }
 }
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,6 +44,9 @@ dependencies {
 
   testImplementation "org.junit.jupiter:junit-jupiter-api:$junitVersion"
   testRuntimeOnly "org.junit.jupiter:junit-jupiter-engine:$junitVersion"
+  testImplementation "org.hamcrest:hamcrest-library:2.2"
+  testImplementation "org.testcontainers:junit-jupiter:1.17.6"
+  testImplementation "org.testcontainers:postgresql:1.17.6"
 }
 
 test {
--- a/app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
+++ b/app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
@@ -1,0 +1,132 @@
+package org.vss.impl.postgres;
+
+import com.google.inject.Inject;
+import com.google.protobuf.ByteString;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import javax.inject.Singleton;
+import org.jooq.DSLContext;
+import org.jooq.Insert;
+import org.jooq.Query;
+import org.jooq.Update;
+import org.vss.GetObjectRequest;
+import org.vss.GetObjectResponse;
+import org.vss.KVStore;
+import org.vss.KeyValue;
+import org.vss.ListKeyVersionsRequest;
+import org.vss.ListKeyVersionsResponse;
+import org.vss.PutObjectRequest;
+import org.vss.PutObjectResponse;
+import org.vss.exception.ConflictException;
+import org.vss.postgres.tables.records.VssDbRecord;
+
+import static org.jooq.impl.DSL.val;
+import static org.vss.postgres.tables.VssDb.VSS_DB;
+
+@Singleton
+public class PostgresBackendImpl implements KVStore {
+
+  private final DSLContext context;
+
+  @Inject
+  public PostgresBackendImpl(DSLContext context) {
+    this.context = context;
+  }
+
+  @Override
+  public GetObjectResponse get(GetObjectRequest request) {
+
+    VssDbRecord vssDbRecord = context.selectFrom(VSS_DB)
+        .where(VSS_DB.STORE_ID.eq(request.getStoreId())
+            .and(VSS_DB.KEY.eq(request.getKey())))
+        .fetchOne();
+
+    final KeyValue keyValue;
+
+    if (vssDbRecord != null) {
+      keyValue = KeyValue.newBuilder()
+          .setKey(vssDbRecord.getKey())
+          .setValue(ByteString.copyFrom(vssDbRecord.getValue()))
+          .setVersion(vssDbRecord.getVersion())
+          .build();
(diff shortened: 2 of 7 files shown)


app/src/main/java/org/vss/impl/postgres/sql/v0_create_vss_db.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE vss_db (
+    store_id character varying(120)  NOT NULL,
```
devrandom, line 2: we probably want primary key constraint on `(store_id, key)`

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -0,0 +1,131 @@
...
+  public PutObjectResponse put(PutObjectRequest request) {
+
+    String storeId = request.getStoreId();
+
+    List<VssDbRecord> vssRecords = new ArrayList<>(request.getTransactionItemsList().stream()
+        .map(kv -> buildVssRecord(storeId, kv)).toList());
+
+    VssDbRecord globalVersionRecord = buildVssRecord(storeId,
+        KeyValue.newBuilder()
+            .setKey(GLOBAL_VERSION_KEY)
```
devrandom, line 71: I think the global version should be stored in a separate table.  otherwise, you have to make sure the caller doesn't try to use this key, and it seems messy.

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -0,0 +1,131 @@
...
+            .setVersion(request.getGlobalVersion())
```
devrandom, line 72: this sets it unconditionally to the user's input and does not handle `null`.  but I thought the logic should be:

- if not specified (i.e. null), increment the current value and don't check it
- if specified, check the current value, abort if different, otherwise increment it

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  private Query buildPutObjectQuery(DSLContext dsl, VssDbRecord vssRecord) {
+    return vssRecord.getVersion() == 0 ? buildInsertRecordQuery(dsl, vssRecord)
+        : buildUpdateRecordQuery(dsl, vssRecord);
+  }
+
+  private Insert<VssDbRecord> buildInsertRecordQuery(DSLContext dsl, VssDbRecord vssRecord) {
+    return dsl.insertInto(VSS_DB)
+        .select(dsl.select(val(vssRecord.getStoreId()), val(vssRecord.getKey()),
```
devrandom, line 102: why not `.values(...)` instead of this subselect?

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -0,0 +1,131 @@
...
+                val(vssRecord.getValue()), val(1))
+            .whereNotExists(dsl.selectOne().from(VSS_DB)
```
devrandom, line 104: I think you can just do `onDuplicateKeyIgnore` here, and the issue will be caught by the batch result check above

app/build.gradle
```diff
@@ -4,13 +4,26 @@ buildscript {
...
     ext.jerseyVersion = '3.1.0'
     ext.junitVersion = '5.9.0'
+    ext.postgresVersion = '42.5.1'
+    ext.jooqVersion = '3.17.4'
+    ext.guiceVersion = '5.1.0'
+    dependencies {
+        classpath  "org.postgresql:postgresql:$postgresVersion"
+    }
+
+    configurations['classpath'].resolutionStrategy.eachDependency {
```
devrandom, line 14: what does this do?

devrandom:
also, would be good to have an integration test.  not sure if it's better to have it against postgres (more precise model of a production environment) or against a lightweight in-memory SQL DB (faster to run the tests).

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
G8XSU: Yes it should be 
* if not specified (i.e. null), do nothing with it, ignore
* if specified, check the current value, abort if different, otherwise increment it

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
G8XSU: subselect was required to use "whereNotExists" but i think we can replace that with "onDuplicateKeyIgnore" and rely on numOfUpdatedRows check

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -66,13 +68,15 @@ public PutObjectResponse put(PutObjectRequest request) {
     List<VssDbRecord> vssRecords = new ArrayList<>(request.getTransactionItemsList().stream()
         .map(kv -> buildVssRecord(storeId, kv)).toList());
 
-    VssDbRecord globalVersionRecord = buildVssRecord(storeId,
-        KeyValue.newBuilder()
-            .setKey(GLOBAL_VERSION_KEY)
-            .setVersion(request.getGlobalVersion())
-            .build());
+    if (request.hasGlobalVersion()) {
```
devrandom, line 71: this doesn't increment the global version when it's not specified in the request.  we always want to increment it, right?

devrandom: I guess we could have the specifications say that it doesn't get incremented if not present in the request, but we should be explicit about it, since the developer might not expect that

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
G8XSU: The advantage is that it can be treated as any other keyvalue and easily handled with exactly same cases. 

Other consideration is, if we want client to be able to override global_version in certain scenarios. My take is we should allow, end of the day its their storage and versioning, they can do whatever they want with it.

There is also case where client just wants to do a "get" on global_version. 
For all purposes, it seems simpler to keep it as normal key-value, otherwise i will need to introduce additional code to support everything.

One case that we will have to handle separately from normal key-value is to not return global_version as part of ListKeyVersions api. (we have dedicated field in response for it)

app/build.gradle
G8XSU: removed in next commit. 
it was required to force dependency resolution to particular version.

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
G8XSU: Based on how optimistic locking works, version needs to be provided by client in order to enact it.
So expectation is that we don't increment it and dont perform the check if developer doesn't supply it. 

That is it is a non-global-version-check-required write. 
We shouldn't be incrementing global_version in this case as client-side will not increment and has no way of knowing without performing a sync.

Based on different application needs, some might not need a global_version check on every write.

G8XSU:
Added an integration test for AbstractKVStore and PostgresIntegration test uses it. 
Can review it as part of this PR or we can separate it out as well in #5

app/src/test/java/org/vss/impl/postgres/PostgresBackendImplIntegrationTest.java
```diff
@@ -0,0 +1,48 @@
...
+  @BeforeEach
+  public void initEach() throws Exception {
+
+    // This is required to get postgres driver in classpath before we attempt to fetch a connection
+    Class.forName("org.postgresql.Driver");
+    Connection conn = DriverManager.getConnection(postgreSQLContainer.getJdbcUrl(),
+        postgreSQLContainer.getUsername(), postgreSQLContainer.getPassword());
+    DSLContext dslContext = DSL.using(conn, SQLDialect.POSTGRES);
+
+    this.kvStore = new PostgresBackendImpl(dslContext);
```
devrandom, line 35: don't we need to close the previous DB connection, otherwise we may run out of file descriptors or such if we have many tests?

G8XSU: Each and every test spins up new db cluster/instance so db connection shouldn't really be a problem. 
But I added a AfterEach block to destroy connection in any case.

For production, we are using connection pool and DSL.using(dataSource, dialect) where jooq/pool does the connection management for us.

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -0,0 +1,133 @@
...
+    if (vssDbRecord != null) {
+      keyValue = KeyValue.newBuilder()
+          .setKey(vssDbRecord.getKey())
+          .setValue(ByteString.copyFrom(vssDbRecord.getValue()))
+          .setVersion(vssDbRecord.getVersion())
+          .build();
+    } else {
+      keyValue = KeyValue.newBuilder()
+          .setKey(request.getKey()).build();
+    }
```
jkczyz, line 56: Should we return an error if there is no record for the key?

jkczyz (commented):
Need to get through the tests still

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
G8XSU: I think an empty response is preferable in this case. 
ResourceNotFound or 404 might represent a myriad of issues and things going wrong. (url wrong etc.)
In this case, we consider it a perfectly valid request to get a key which does not exist or to check existence of a key.
I see no harm in mixing keys that exist with no data and keys which don't exist, to make clients life easier instead of throwing an exception in one case.

Note: This is only applicable if client has permission to storeId, if not then we would want to throw ResourceNotFound.

However, this is a controversial topic and there are reasons to go either way.

app/src/test/java/org/vss/impl/postgres/PostgresBackendImplIntegrationTest.java
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  @AfterEach
+  void destroy() throws Exception {
+    this.connection.close();
+  }
+
+  private void createTable(DSLContext dslContext) {
+    dslContext.execute("CREATE TABLE vss_db ("
+        + "store_id character varying(120)  NOT NULL CHECK (store_id <> ''),"
```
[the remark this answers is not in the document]
G8XSU: will fix this.

app/src/test/java/org/vss/impl/postgres/PostgresBackendImplIntegrationTest.java
```diff
@@ -0,0 +1,56 @@
...
+        + "key character varying(120)  NOT NULL,"
+        + "value bytea  NULL,"
+        + "version bigint  NOT NULL,"
+        + "PRIMARY KEY (store_id, key));");
```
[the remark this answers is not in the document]
G8XSU: Note that the sql file is present only for documentation purpose and git history. 
Its not going to be directly used in db creation, will need to be done manually. 
I wanted to avoid a filepath dependency in tests, let me know your thoughts.

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -0,0 +1,133 @@
...
+
+      int[] batchResult = dsl.batch(batchQueries).execute();
+
+      for (int numOfRowsUpdated : batchResult) {
+        if (numOfRowsUpdated == 0) {
+          throw new ConflictException(
+              "Transaction could not be completed due to a possible conflict");
+        }
+      }
+    });
```
jkczyz, line 94: Could you test the case where one key failing causes the entire transaction to fail? (i.e., the successful key is not updated)

app/src/main/java/org/vss/impl/postgres/sql/v0_create_vss_db.sql
```diff
@@ -1,8 +1,7 @@
...
-    value bytea  NULL,
-    version bigint  NOT NULL,
+    store_id character varying(120) NOT NULL CHECK (store_id <> ''),
+    key character varying(120) NOT NULL,
+    value bytea NULL,
+    version bigint NOT NULL,
     PRIMARY KEY (store_id, key)
-);
-
+);
```
jkczyz, line 7: Add newline.

jkczyz (commented):
Looks like some of the comments for the first commit were resolved in the second commit. You'll want to make sure they are resolved in the right commit so that they are self-contained. Otherwise, looks good and sorry about the delay.

app/src/main/java/org/vss/impl/postgres/sql/v0_create_vss_db.sql
G8XSU: I earlier got comment about removing newlines/blanks at end of files, 
bit confused, let me know if i misunderstood.

jkczyz: Sorry, please disregard. I thought that Github was showing the "No newline at end of file" symbol, but I guess I was mistaken.

G8XSU: Ok, i might have fixed it, 
So just to clarify there "should" be a newline at EOF? 

Addressed other comments as well.

jkczyz: Correct, it should contain a newline, just not an empty line.

jkczyz (approved):

PR #6, opened 2023-01-25
Add AbstractKVStoreIntegrationTest and PostgresImpl Test

G8XSU:
Based on #5

PR #8, opened 2023-04-19
Add Implementation for ListKeyVersions Api and AbstractKVStore tests for the same

G8XSU:
* Add ListKeyVersions Api Implementation
* Add ListKeyVersions Api AbstractKVStore Integration Tests

Depends on #5 and #6

### DIFF
--- a/app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
+++ b/app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
@@ -5,6 +5,7 @@ import com.google.protobuf.ByteString;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import javax.inject.Singleton;
 import org.jooq.DSLContext;
 import org.jooq.Insert;
@@ -21,12 +22,12 @@ import org.vss.PutObjectResponse;
 import org.vss.exception.ConflictException;
 import org.vss.postgres.tables.records.VssDbRecord;
 
-import static org.jooq.impl.DSL.val;
 import static org.vss.postgres.tables.VssDb.VSS_DB;
 
 @Singleton
 public class PostgresBackendImpl implements KVStore {
 
+  private static final int LIST_KEY_VERSIONS_MAX_PAGE_SIZE = 100;
   private final DSLContext context;
 
   @Inject
@@ -127,6 +128,54 @@ public class PostgresBackendImpl implements KVStore {
 
   @Override
   public ListKeyVersionsResponse listKeyVersions(ListKeyVersionsRequest request) {
-    throw new UnsupportedOperationException("Operation not implemented");
+    String storeId = request.getStoreId();
+    String keyPrefix = request.getKeyPrefix();
+    String pageToken = request.getPageToken();
+    int pageSize = request.hasPageSize() ? request.getPageSize() : Integer.MAX_VALUE;
+
+    // Only fetch global_version for first page.
+    // Fetch global_version before fetching any key_versions to ensure that,
+    // all current key_versions were stored at global_version or later.
+    Long globalVersion = null;
+    if (!request.hasPageToken()) {
+      GetObjectRequest getGlobalVersionRequest = GetObjectRequest.newBuilder()
+          .setStoreId(storeId)
+          .setKey(GLOBAL_VERSION_KEY)
+          .build();
+      globalVersion = get(getGlobalVersionRequest).getValue().getVersion();
+    }
+
+    List<VssDbRecord> vssDbRecords = context.select(VSS_DB.KEY, VSS_DB.VERSION).from(VSS_DB)
+        .where(VSS_DB.STORE_ID.eq(storeId)
+            .and(VSS_DB.KEY.startsWith(keyPrefix)))
+        .orderBy(VSS_DB.KEY)
+        .seek(pageToken)
+        .limit(Math.min(pageSize, LIST_KEY_VERSIONS_MAX_PAGE_SIZE))
+        .stream()
+        .map(record -> record.into(VssDbRecord.class))
+        .toList();
+
+    List<KeyValue> keyVersions = vssDbRecords.stream()
+        .filter(kv -> !GLOBAL_VERSION_KEY.equals(kv.getKey()))
+        .map(kv -> KeyValue.newBuilder()
+            .setKey(kv.getKey())
+            .setVersion(kv.getVersion())
+            .build())
+        .toList();
+
+    String nextPageToken = "";
+    if (!keyVersions.isEmpty()) {
+      nextPageToken = keyVersions.get(keyVersions.size() - 1).getKey();
+    }
+
+    ListKeyVersionsResponse.Builder responseBuilder = ListKeyVersionsResponse.newBuilder()
+        .addAllKeyVersions(keyVersions)
+        .setNextPageToken(nextPageToken);
+
+    if (Objects.nonNull(globalVersion)) {
+      responseBuilder.setGlobalVersion(globalVersion);
+    }
+
+    return responseBuilder.build();
   }
 }
--- a/app/src/test/java/org/vss/AbstractKVStoreIntegrationTest.java
+++ b/app/src/test/java/org/vss/AbstractKVStoreIntegrationTest.java
@@ -2,14 +2,22 @@ package org.vss;
 
 import com.google.protobuf.ByteString;
 import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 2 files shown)


app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -148,8 +151,8 @@ public ListKeyVersionsResponse listKeyVersions(ListKeyVersionsRequest request) {
             .and(VSS_DB.KEY.startsWith(keyPrefix)))
         .orderBy(VSS_DB.KEY)
         .seek(pageToken)
-        .limit(Math.min(pageSize, 50))
-        .getResult().map(record -> record.into(VssDbRecord.class));
+        .limit(Math.min(pageSize, LIST_KEY_VERSIONS_MAX_PAGE_SIZE))
+        .stream().map(record -> record.into(VssDbRecord.class)).toList();
```
jkczyz, line 155: nit: put `toList();` on the next line

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -164,10 +167,14 @@ public ListKeyVersionsResponse listKeyVersions(ListKeyVersionsRequest request) {
...
-        .setGlobalVersion(globalVersion)
+    ListKeyVersionsResponse.Builder responseBuilder = ListKeyVersionsResponse.newBuilder()
         .addAllKeyVersions(keyVersions)
-        .setNextPageToken(nextPageToken)
-        .build();
+        .setNextPageToken(nextPageToken);
+
+    if (Objects.nonNull(globalVersion)) {
+      responseBuilder.setGlobalVersion(globalVersion);
+    }
```
jkczyz, line 176: I'm forgetting where we landed on this. How does the client interpret this value?

app/src/test/java/org/vss/AbstractKVStoreIntegrationTest.java
```diff
@@ -163,6 +179,177 @@ void getShouldReturnCorrectValueWhenKeyExists() {
...
+    putObjects(1002L, List.of(kv("k2", "k2v3", 2)));
+
+    ListKeyVersionsResponse previousPage = null;
+    List<KeyValue> allKeyVersions = new ArrayList<>();
+
+    while (previousPage == null || !previousPage.getKeyVersionsList().isEmpty()) {
+      ListKeyVersionsResponse currentPage;
+
+      if (previousPage == null) {
+        currentPage = list(null, null, null);
```
jkczyz, line 201: I'd find this more readable if `list` took a builder.

app/src/test/java/org/vss/AbstractKVStoreIntegrationTest.java
```diff
@@ -163,6 +179,177 @@ void getShouldReturnCorrectValueWhenKeyExists() {
...
+
+  @Test
+  void listShouldLimitMaxPageSize() {
+
+    int totalKvObjects = 10000;
+
+    // Each implementation is free to choose its own max_page_size but there should be a reasonable max
+    // keeping scalability and performance in mind.
+    // Revisit this test case if some implementation wants to support higher page size.
+    int vssArbitraryPageSizeMax = 3000;
```
jkczyz, line 324: Not sure I follow the reason for this test. Isn't there already a lower constant in the code?

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -127,6 +126,48 @@ private VssDbRecord buildVssRecord(String storeId, KeyValue kv) {
...
+    List<VssDbRecord> vssDbRecords = context.select(VSS_DB.KEY, VSS_DB.VERSION).from(VSS_DB)
+        .where(VSS_DB.STORE_ID.eq(storeId)
+            .and(VSS_DB.KEY.startsWith(keyPrefix)))
+        .orderBy(VSS_DB.KEY)
+        .seek(pageToken)
+        .limit(Math.min(pageSize, 50))
+        .getResult().map(record -> record.into(VssDbRecord.class));
+
+    List<KeyValue> keyVersions = vssDbRecords.stream()
+        .filter(kv -> !GLOBAL_VERSION_KEY.equals(kv.getKey()))
```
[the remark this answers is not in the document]
G8XSU: Yes, it can mean number of entries in response can be one less than pageSize. 
But that shouldn't be a concern and client should not assume response to contain specific number of entries. 
For e.g. max number of results in paginated response can change at anytime with no notice to client.

We already caution against this in api doc in proto: 
"Caution: Clients must not assume a specific number of key_versions to be present in a page for paginated response."
Only way to know whether nextPage exists or not is by presence of nextPageToken.

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
G8XSU: Since clients need globalVersion to make subsequent successful writes, they need to update their client-side global_version for it.
Quoting: 
```
  // global_version is only returned in response for the first page of the ListKeyVersionsResponse
  // and is guaranteed to be read before reading any key-versions.
  //
  // In case of refreshing the complete key-version view on the client-side, correct usage for
  // the returned global_version is as following:
  //   1. Read global_version from the first page of paginated response and save it as local variable.
  //   2. Update all the key_versions on client-side from all the pages of paginated response.
  //   3. Update global_version on client_side from the local variable saved in step-1.
  // This ensures that on client-side, all current key_versions were stored at global_version or later.
  // This guarantee is helpful for ensuring the versioning correctness if using the global_version
  // in PutObject API and can help avoid the race conditions related to it.
```

app/src/test/java/org/vss/AbstractKVStoreIntegrationTest.java
G8XSU: listRequest follows a builder pattern, but this particular "list" function is just for testcase to avoid making them too verbose. Whenever we are using any argument, we extract it to variable and use it for readability.

app/src/test/java/org/vss/AbstractKVStoreIntegrationTest.java
G8XSU: MAX_PAGE_SIZE is implementation specific. 
Yes, there is a lower page_size(100) currently configured for postgres implementation but its not going to be same for all impl.

Since these tests lie in AbstractKVStore, we just test that there is some limit to page size from implementation and the behaviour in case cllient requests without page size or with page_size too large.

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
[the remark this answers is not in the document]
jkczyz: That doesn't answer my second question. :)

> Should we filter at the SQL level instead?

Is there a reason not to?

app/src/test/java/org/vss/AbstractKVStoreIntegrationTest.java
G8XSU: i earlier had that builder everywhere, seems more readable like this 😅

jkczyz: Yeah, feel free to keep as is. Could possibly be a little more readable if there were special case functions that called `list` to avoid all the `null`s. But I'll leave that decision to you.

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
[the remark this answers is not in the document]
G8XSU: There is no big reason in case of SQL, i can do that. (apart from client/api-expectation and precedent) 
It is just something to keep in mind that this operation might not be supported by all KV-database i.e. (list along with key not equals).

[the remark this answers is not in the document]
jkczyz: Thanks ok. Feel fee to leave it as is.

jkczyz:
Could you rebase?

G8XSU:
Done

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -138,10 +138,11 @@ public ListKeyVersionsResponse listKeyVersions(ListKeyVersionsRequest request) {
     // all current key_versions were stored at global_version or later.
     Long globalVersion = null;
     if (!request.hasPageToken()) {
-      globalVersion = get(GetObjectRequest.newBuilder()
+      GetObjectRequest getGlobalVersionRequest = GetObjectRequest.newBuilder()
           .setStoreId(storeId)
           .setKey(GLOBAL_VERSION_KEY)
-          .build()).getValue().getVersion();
+          .build();
+      globalVersion = get(getGlobalVersionRequest).getValue().getVersion();
```
jkczyz: Changes here a below belong on the first commit.

G8XSU: fixed it 👍

PR #10, opened 2023-05-12
Add AbstractVssApi, get/put/list api handlers

G8XSU:
Add AbstractVssApi, get/put/list api handlers

PR #12, opened 2023-05-25
Add GetObjectApiTest, PutObjectsApiTest, ListKeyVersionsApiTest

G8XSU:
Depends on #10

### DIFF
--- a/app/src/test/java/org/vss/api/GetObjectApiTest.java
+++ b/app/src/test/java/org/vss/api/GetObjectApiTest.java
@@ -1,0 +1,86 @@
+package org.vss.api;
+
+import com.google.protobuf.ByteString;
+import jakarta.ws.rs.core.Response;
+import java.nio.charset.StandardCharsets;
+import java.util.stream.Stream;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.vss.ErrorCode;
+import org.vss.ErrorResponse;
+import org.vss.GetObjectRequest;
+import org.vss.GetObjectResponse;
+import org.vss.KVStore;
+import org.vss.KeyValue;
+import org.vss.exception.ConflictException;
+
+import static org.hamcrest.MatcherAssert.assertThat;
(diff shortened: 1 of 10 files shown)


app/src/test/java/org/vss/api/GetObjectApiTest.java
```diff
@@ -0,0 +1,87 @@
...
+      ByteString.copyFrom("test_value", StandardCharsets.UTF_8)).build();
+
+  @BeforeEach
+  void setUp() {
+    mockKVStore = mock(KVStore.class);
+    getObjectApi = new GetObjectApi(mockKVStore);
+  }
+
+  @Test
+  void execute_ValidPayload_ReturnsResponse() {
```
G8XSU, line 43: i tried creating a common base testcase but it was more complex and less readable since many api specific calls are interleaved.

G8XSU: these are the abstract methods a sub-test would have had to implement in that case, and it seemed difficult to understand test-case.
```
  abstract void beforeEachSetup();
  
  abstract void verifyAfterTest();

  abstract GeneratedMessageV3 createExpectedRequest();

  abstract GeneratedMessageV3 createExpectedResponse();
  
  abstract Response executeValidRequest();

  abstract Response executeErroringRequest(Class<? extends Exception> exceptionClass)
      throws Exception;
      
```

app/src/test/java/org/vss/api/GetObjectApiTest.java
```diff
@@ -0,0 +1,87 @@
...
+  @MethodSource("provideErrorTestCases")
+  void execute_InvalidPayload_ReturnsErrorResponse(Class<? extends Exception> exceptionClass,
+      ErrorCode errorCode) throws Exception {
+    GetObjectRequest expectedRequest = GetObjectRequest.newBuilder()
+        .setStoreId(TEST_STORE_ID)
+        .setKey(TEST_KEY)
+        .build();
+    byte[] payload = expectedRequest.toByteArray();
+    when(mockKVStore.get(any())).thenThrow(
+        exceptionClass.getDeclaredConstructor(String.class).newInstance("errorMsg"));
```
jkczyz, line 67: Wouldn't be simpler to pass in the constructed exception?

G8XSU:
Hi @jkczyz
Will you be approving #10 separately or should i merge just this combined PR ?

PR #14, opened 2023-06-15
Center align image and fix loop in image

G8XSU:
* Center align image
* fix previous render which missed a loop in image


moneyball (approved):
visually confirmed tACK

PR #15, opened 2023-07-25
Add Delete functionality

G8XSU:
We add delete functionality as both:
* Transaction support in PutObject Api, which allows simultaneous delete of keys while writing other keys. This is necessary in order to support CRUD transactions.
* An idempotent api to allow for applications which can safely merge deletes and don't require transactions with put&delete.

This follows similar pattern followed in popular KVStore Databases, such as aws-dynamodb which allows for both delete api and transaction write api.

For now, ignore the naming for "PutObject" API, will rename it to WriteObject in separate rename-commit.

### DIFF
--- a/app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
+++ b/app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
@@ -11,6 +11,8 @@ import org.jooq.DSLContext;
 import org.jooq.Insert;
 import org.jooq.Query;
 import org.jooq.Update;
+import org.vss.DeleteObjectRequest;
+import org.vss.DeleteObjectResponse;
 import org.vss.GetObjectRequest;
 import org.vss.GetObjectResponse;
 import org.vss.KVStore;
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -66,23 +66,31 @@ public PutObjectResponse put(PutObjectRequest request) {
...
     context.transaction((ctx) -> {
       DSLContext dsl = ctx.dsl();
-      List<Query> batchQueries = vssRecords.stream()
-          .map(vssRecord -> buildPutObjectQuery(dsl, vssRecord)).toList();
+      List<Query> batchQueries = new ArrayList<>();
+
+      batchQueries.addAll(vssPutRecords.stream()
+          .map(vssRecord -> buildPutObjectQuery(dsl, vssRecord)).toList());
+      batchQueries.addAll(vssDeleteRecords.stream()
+          .map(vssRecord -> buildDeleteObjectQuery(dsl, vssRecord)).toList());
```
jkczyz, line 95: Does there need to be a check for distinct keys across both lists?

jkczyz (commented):
Concept ACK.

jkczyz (approved):

PR #17, opened 2023-08-18
Add Non-conditional Put/Delete functionality

G8XSU:
Similar to popular kvstores, we should also support non-conditional put/delete operations. This opens up new patterns for usage of VSS, allows clients to easily onboard to VSS (without versioning, just for remote-backups) or to just use global_version for versioning and avoid key-level version checks.

DynamoDB: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html
If-match in cosmos: https://learn.microsoft.com/en-us/rest/api/cosmos-db/replace-a-document

### DIFF
--- a/app/src/main/proto/vss.proto
+++ b/app/src/main/proto/vss.proto
@@ -73,18 +73,26 @@ message PutObjectRequest {
   // a database-transaction in an all-or-nothing fashion.
   // All Items in a single `PutObjectRequest` must have distinct keys.
   //
-  // Clients are expected to store a `version` against every `key`.
-  // The write will succeed if the current DB version against the `key` is the same as in the request.
-  // When initiating a `PutObjectRequest`, the request should contain their client-side version for
-  // that key-value.
+  // Key-level versioning (Conditional Write):
+  //   Clients are expected to store a `version` against every `key`.
+  //   The write will succeed if the current DB version against the `key` is the same as in the request.
+  //   When initiating a `PutObjectRequest`, the request should contain their client-side `version`
+  //   for that key-value.
   //
-  // For the first write of any key, the `version` should be '0'. If the write succeeds, the client
-  // must increment their corresponding key versions (client-side) by 1.
-  // The server increments key versions (server-side) for every successful write, hence this
-  // client-side increment is required to ensure matching versions. These updated key versions should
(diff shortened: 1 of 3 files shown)


jkczyz (commented):
How will this be exposed in the clients? Is there risk of accidentally corrupting data if a device has misconfigured its client to use non-conditional writes while another device is properly configured to do conditional writes? Is there any way to guard agains this?

app/src/main/proto/vss.proto
```diff
@@ -73,18 +73,26 @@ message PutObjectRequest {
...
+  //   client-side increment is required to ensure matching versions. These updated key versions should
+  //   be used in subsequent `PutObjectRequest`s for the keys.
+  //
+  //   Requests with a conflicting/mismatched version will fail with `CONFLICT_EXCEPTION` as ErrorCode
+  //   for conditional writes.
+  //
+  // Skipping key-level versioning (Non-conditional Write):
+  //   If you wish to skip key-level version checks, set the `version` against the `key` to '-1'.
+  //   This will perform a non-conditional upsert query. After any non-conditional upsert, the
+  //   `version` against the `key` is reset to '1'. Hence the next `PutObjectRequest` for the `key`
```
jkczyz, line 94: nit: Merge these two sentences as:

> This will perform a non-conditional upsert query, after which the `version` against the `key` is reset to '1'.

app/src/main/proto/vss.proto
```diff
@@ -101,13 +109,20 @@ message PutObjectRequest {
...
+  //   the `version` specified in the request.
+  //
+  // Skipping key-level versioning (Non-conditional Delete):
+  //   If you wish to skip key-level version checks, set the `version` against the `key` to '-1'.
+  //   This will perform a non-conditional delete query.
   //
   // Fails with `CONFLICT_EXCEPTION` as the ErrorCode if:
   //   * The requested item does not exist.
-  //   * The requested item does exist but there is a version-number mismatch with the one in the database.
+  //   * The requested item does exist but there is a version-number mismatch(in conditional delete)
```
jkczyz, line 124: Add a space before `(`

jkczyz (approved):
Just a few minor comments. Otherwise, looks good.

PR #19, opened 2023-08-24
Upgrade deps to support gradle-8.0

G8XSU:
Gradle 8.0 converts some warnings into errors compared to 7.4.2. 

With this change, build should be supported on both.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,6 +1,5 @@
 buildscript {
-  ext.gradleVersion = '7.5.1'
-  ext.protobufPlugInVersion = '0.8.12'
+  ext.protobufPlugInVersion = '0.9.4'
   ext.protobufVersion = '3.21.7'
   ext.jerseyVersion = '3.1.0'
   ext.junitVersion = '5.9.0'


app/src/test/java/org/vss/AbstractKVStoreIntegrationTest.java
```diff
@@ -46,7 +46,7 @@ void putShouldSucceedWhenSingleObjectPutOperation() {
     response = getObject("k2");
     assertThat(response.getKey(), is("k2"));
     assertThat(response.getVersion(), is(1L));
-    assertThat(response.getValue().toStringUtf8(), is("k1v3"));
+    assertThat(response.getValue().toStringUtf8(), is("k2v3"));
```
jkczyz, line 49: Ah, should have caught this before... 😅 Do we have plans to set up CI testing?

app/build.gradle
```diff
@@ -1,6 +1,5 @@
 buildscript {
-  ext.gradleVersion = '7.5.1'
```
jkczyz, line 2: Does removing this mean we get the most recent version? Should we fix it to 8.0 instead?

jkczyz (approved):

PR #20, opened 2023-08-31
Add CI to check E2E Build & Deploy pipeline

G8XSU:
* Add CI
* Change default postgres password, since default postgres docker image refuses to start with empty password

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -1,0 +1,80 @@
+name: Continuous Integration
+
+on: [push, pull_request]
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+
+    services:
+      postgres:
+        image: postgres:15
+        ports:
+          - 5432:5432
+        env:
+          POSTGRES_DB: postgres
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: YOU_MUST_CHANGE_THIS_PASSWORD
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Java
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'corretto'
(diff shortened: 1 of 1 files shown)


app/build.gradle
```diff
@@ -79,7 +79,7 @@ jooq {
           driver = 'org.postgresql.Driver'
           url = 'jdbc:postgresql://localhost:5432/postgres'
           user = 'postgres'
-          password = ''
+          password = 'testpassword'
```
jkczyz, line 82: What's the relationship between this file and `hikariJdbc.properties`. Can this file reference properties from the latter?

app/src/main/resources/hikariJdbc.properties
```diff
@@ -2,7 +2,7 @@
 
 jdbc.url=jdbc:postgresql://localhost:5432/postgres
 jdbc.username=postgres
-jdbc.password=
+jdbc.password=testpassword
```
jkczyz, line 5: Are users expected to change this? If so, could we name it such that this is obvious? Isn't clear to me if this has something to do with testing.

.github/workflows/build-and-deploy.yml
```diff
@@ -0,0 +1,82 @@
...
+          ./gradlew build
+
+          docker cp app/build/libs/app-1.0.war tomcat:/usr/local/tomcat/webapps/vss.war
+
+      - name: Hit endpoint to verify service is up
+        run: |
+          # Wait for tomcat-startup
+          sleep 5
+
+          # Verify curl exits with error-code 0
```
jkczyz, line 70: Isn't clear what this comment is referring to. I'd just drop it if a non-zero exit causes a failure.

app/src/main/resources/hikariJdbc.properties
jkczyz: How about "YOU_MUST_CHANGE_THIS_PASSWORD" or something as obvious?

app/build.gradle
```diff
@@ -79,7 +79,7 @@ jooq {
           driver = 'org.postgresql.Driver'
           url = 'jdbc:postgresql://localhost:5432/postgres'
           user = 'postgres'
-          password = 'testpassword'
+          password = 'YOU_MUST_CHANGE_THIS_PASSWORD'
```
jkczyz, line 82: Could you put both password changes in the first commit? Otherwise, LGTM.

jkczyz (approved):

PR #21, opened 2023-09-13
Throw NoSuchKeyException when key doesn't exist in get

G8XSU:
Naming for error code matches aws s3 and gcp object store, seems like popular enough convention. 

Even though there are some popular KVStore's which do not throw such exception on keyNotFound such as Aws-DDB (arguably the most popular one), we can still figure out that key didn't exist if value was returned as null or by similar means in other stores. 

Motivation: Such checks are best performed in backend than having client checking empty or null value.
This is also in-line with more critical requirement for functioning of NotFound introduced by MonitorUpdatingPersister Design.

### DIFF
--- a/app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
+++ b/app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
@@ -23,6 +23,7 @@ import org.vss.ListKeyVersionsResponse;
 import org.vss.PutObjectRequest;
 import org.vss.PutObjectResponse;
 import org.vss.exception.ConflictException;
+import org.vss.exception.NoSuchKeyException;
 import org.vss.postgres.tables.records.VssDbRecord;
 
 import static org.vss.postgres.tables.VssDb.VSS_DB;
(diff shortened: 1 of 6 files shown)


app/src/main/proto/vss.proto
```diff
@@ -276,6 +279,9 @@ enum ErrorCode {
   // An internal server error occurred, client is probably at no fault and can safely retry this
   // error with exponential backoff.
   INTERNAL_SERVER_EXCEPTION = 3;
+
+  // NO_SUCH_KEY_EXCEPTION is used when specified `key` in `GetObjectRequest` does not exist.
```
jkczyz, line 283: s/specified/the specified
s/in/in a

app/src/main/java/org/vss/impl/postgres/PostgresBackendImpl.java
```diff
@@ -47,16 +53,19 @@ public GetObjectResponse get(GetObjectRequest request) {
...
 
     final KeyValue keyValue;
-
-    if (vssDbRecord != null) {
+    if (vssDbRecord == null) {
+      if (GLOBAL_VERSION_KEY.equals(request.getKey())) {
+        keyValue = DEFAULT_GLOBAL_VERSION_KV;
+      } else {
+        throw new NoSuchKeyException(
+            "Specified key:" + request.getKey() + " in request does not exist.");
```
jkczyz, line 61: Space needed after colon... probably no colon needed actually or use parentheses instead.

jkczyz (commented):
Could you update the commit message to contain the "what" and "why" from the PR summary? git, not github, should be our source of truth.

jkczyz (approved):

PR #22, opened 2023-09-13
Add Storable Helper Object

G8XSU:
Motivation: Provides helper object for client to use, avoids some foot-guns in client-side encryption and ensures backward compatibility.
We don't want to burden client to think about backward compatibility and version-integrity while deciding which fields to serialize & store inside the value for put. 
This is kind of part of vss-protocol.

### DIFF
--- a/app/src/main/proto/vss.proto
+++ b/app/src/main/proto/vss.proto
@@ -293,7 +293,48 @@ message KeyValue {
   int64 version = 2;
 
   // Object value in bytes which is stored (in put) and fetched (in get).
-  // Clients must encrypt this blob client-side before sending it over the wire to server in order
-  // to preserve privacy and security.
+  // Clients must encrypt the secret contents of this blob client-side before sending it over the
+  // wire to the server in order to preserve privacy and security.
+  // Clients may use a `Storable` object, serialize it and set it here.
   bytes value = 3;
+}
+
+// Represents a storable object that can be serialized and stored as `value` in `PutObjectRequest`.
+// Only provided as a helper object for ease of use by clients.
+// Clients MUST encrypt the `PlaintextBlob` before using it as `data` in `Storable`.
+// The server does not use or read anything from `Storable`, Clients may use its fields as
+// required.
+message Storable {
+
+  // Represents an encrypted and serialized `PlaintextBlob`. MUST encrypt the whole `PlaintextBlob`
+  // using client-side encryption before setting here.
+  bytes data = 1;
+
+  // Represents encryption related metadata
+  EncryptionMetadata encryption_metadata = 2;
+}
+
+// Represents encryption related metadata
+message EncryptionMetadata {
+  // The encryption algorithm used for encrypting the `PlaintextBlob`.
+  string cipher_format = 1;
+
+  // The nonce used for encryption. Nonce is a random or unique value used to ensure that the same
+  // plaintext results in different ciphertexts every time it is encrypted.
+  bytes nonce = 2;
+
+  // The authentication tag used for encryption. It provides integrity and authenticity assurance
+  // for the encrypted data.
+  bytes tag = 3;
+}
+
+// Represents a data blob, which is encrypted, serialized and later used in `Storable.data`.
+// Since the whole `Storable.data` is client-side encrypted, the server cannot understand this.
+message PlaintextBlob {
+
+  // The unencrypted value.
+  bytes value = 1;
+
+  // The version of the value. Can be used by client to verify version integrity.
+  int64 version = 2;
 }


jkczyz:
How would this be used client-side? Why do we need to use a protocol buffer for it?

app/src/main/proto/vss.proto
```diff
@@ -293,7 +293,40 @@ message KeyValue {
   int64 version = 2;
 
   // Object value in bytes which is stored (in put) and fetched (in get).
-  // Clients must encrypt this blob client-side before sending it over the wire to server in order
-  // to preserve privacy and security.
+  // Clients must encrypt secret contents of this blob client-side before sending it over the wire
+  // to server in order to preserve privacy and security.
```
jkczyz, line 297: s/server/the server

app/src/main/proto/vss.proto
```diff
@@ -293,7 +293,40 @@ message KeyValue {
...
+  // Clients may use a `Storable` object on client-side and set it here.
```
jkczyz, line 298: "client-side" seems redundant given the subject of the sentenced is "Clients". Instead, phrase it something like serializing the object before sending it to the server.

app/src/main/proto/vss.proto
```diff
@@ -293,7 +293,40 @@ message KeyValue {
...
   bytes value = 3;
 }
+
+// Represents a storable object that can be serialized and stored as value in `PutObjectRequest`.
+// Server does not use or read anything from Storable, clients may use fields as required.
```
jkczyz, line 303: s/Server/The server
s/, clients/Clients
s/fields/its fields

jkczyz, line 303: Would put the part about clients first. Then say "However, the server ...".

app/src/main/proto/vss.proto
```diff
@@ -293,7 +293,40 @@ message KeyValue {
...
+// Only provided as a helper object for ease of use by clients.
+message Storable {
+
+  // Represents an encrypted and serialized `CipherBlob`. MUST encrypt the whole `CipherBlob` using
+  // client-side encryption before setting here.
+  bytes cipher_blob = 1;
+
+  // The encryption algorithm used for encryption of the `cipher_blob`.
```
jkczyz, line 311: s/encryption of/encrypting

app/src/main/proto/vss.proto
```diff
@@ -293,7 +293,40 @@ message KeyValue {
...
+  // plaintext results in different ciphertexts every time it is encrypted.
+  bytes nonce = 3;
+
+  // The authentication tag used for encryption. It provides integrity and authenticity assurance
+  // for the encrypted data.
+  bytes tag = 4;
+}
+
+// Represents a data blob, which is encrypted and serialized and later set in `Storable.cipher_blob`.
+// Since whole CipherBlob is client-side encrypted, server cannot understand this.
```
jkczyz, line 324: s/server/the server

jkczyz (commented):
Just a bunch of nits and naming concerns. Otherwise, looks good.

app/src/main/proto/vss.proto
```diff
@@ -293,7 +293,40 @@ message KeyValue {
...
+  // Clients must encrypt secret contents of this blob client-side before sending it over the wire
+  // to server in order to preserve privacy and security.
+  // Clients may use a `Storable` object on client-side and set it here.
   bytes value = 3;
 }
+
+// Represents a storable object that can be serialized and stored as value in `PutObjectRequest`.
+// Server does not use or read anything from Storable, clients may use fields as required.
+// Only provided as a helper object for ease of use by clients.
+message Storable {
```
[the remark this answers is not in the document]
jkczyz: Sure, but it's a data abstraction representing encrypted data. The fact that it also contains metadata doesn't change that. The name `Storable` can describe anything and could have just as easily been named `Data`.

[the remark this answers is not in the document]
jkczyz: > Storable is something that will be stored, whereas data is very non-specific. (it is a naming convention for objects that are to be stored in database, [link](https://docs.oracle.com/en/industries/communications/billing-revenue/12.0/dev-guide/understanding-storable-classes1.html#GUID-A165AD0B-E419-4ACA-936C-DB053BED819B))
> 

What would you name a message that contained non-encyrpted data to store? This message precludes using it for that but is named otherwise.

The example you cite is used to map class fields to database columns. This doesn't have a similar usage.

> * I refrain from naming it encryptedData because it is not encrypted. (could be confusing)

I'm confused now... Doesn't it contain encrypted data and related metadata?

> Also, encryption is not the only characteristic of inner_data just one of them, for example if we add compression. We don't want to add EncryptedCompressedData.

If you want to call it `Storable`, I feel that the value field should have a generic name and any other optional fields for metadata should be separate messages with their own set of fields (e.g., an `encryption` field holding an `EncryptionMetadata` message). Then a reader would examine the metadata fields in order to perform operations (e.g., decrypt, uncompress) on the value.

app/src/main/proto/vss.proto
```diff
@@ -293,7 +293,42 @@ message KeyValue {
...
+// Represents a storable object that can be serialized and stored as value in `PutObjectRequest`.
+// Only provided as a helper object for ease of use by clients.
+// Clients MUST encrypt the `PlaintextBlob` before using it as `value` in `Storable`.
+// The server does not use or read anything from `Storable`, Clients may use its fields as
+// required.
+message Storable {
+
+  // Represents an encrypted and serialized `PlaintextBlob`. MUST encrypt the whole `PlaintextBlob`
+  // using client-side encryption before setting here.
+  bytes value = 1;
```
jkczyz, line 311: Maybe `data` would be better if this is intended to store an encrypted `PlaintextBlob`, which is composed of a value and version?

app/src/main/proto/vss.proto
```diff
@@ -293,7 +293,42 @@ message KeyValue {
...
+}
+
+// Represents a data blob, which is encrypted, serialized and later used in `Storable.value`.
+// Since the whole `Storable.value` is client-side encrypted, the server cannot understand this.
+message PlaintextBlob {
+
+  // The decrypted data blob.
+  bytes value = 1;
+
+  // The version of the data blob. Can be used by client to verify version integrity.
```
jkczyz, line 332: s/data blob/`value`

app/src/main/proto/vss.proto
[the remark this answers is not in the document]
jkczyz: > I don't understand this recommendation fully but feel this is overly complex.
> 
> Will try to discuss with you offline.

If you use optional message fields, you can support cases where the data is encrypted but is not compressed, for instance. Then the client library can use existence checks to decrypt, uncompress, etc.

app/src/main/proto/vss.proto
```diff
@@ -293,7 +293,42 @@ message KeyValue {
...
+  // The authentication tag used for encryption. It provides integrity and authenticity assurance
+  // for the encrypted data.
+  bytes tag = 4;
+}
+
+// Represents a data blob, which is encrypted, serialized and later used in `Storable.data`.
+// Since the whole `Storable.data` is client-side encrypted, the server cannot understand this.
+message PlaintextBlob {
+
+  // The decrypted value.
```
jkczyz, line 329: s/decrypted/unencrypted

app/src/main/proto/vss.proto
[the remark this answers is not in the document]
jkczyz: Thanks, but looks like this wasn't addressed. The idea is to have one optional message field using a new message as the type for the encryption metadata fields. The client code for encryption would return this to set on the `Storable`.

Then if compression is added later, another message for it's metadata would be added using a separate field. And the client code for compression would likewise return its own metadata message to have set on `Storable`.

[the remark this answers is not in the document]
jkczyz: Yeah, I'd prefer a separate message as it can define everything that must be set. With separate fields, you can have a weird state where only some of the fields are set even if all of them must be. Or if none of them should be set, you'd have to give them default values, in Rust, I'd imagine? How do optional fields work in the Rust API, come to think of it? Do they use `Option`?

It can also be initialized inline when constructing the `Storable`, so I don't think it's much boilerplate.

```rust
Storable {
	value: cipher_blob,
	encryption_metadata: EncryptionMetadata {
		nonce: nonce.to_vec(),
		tag: tag.to_vec(),
		cipher_format: CHACHA20_CIPHER_NAME.to_string(),
	},
}
```

jkczyz (approved):

PR #25, opened 2023-09-26
Allow Longer keys upto 600 chars

G8XSU:
Motivation: Earlier we were conservative about our key-size since it is easier to increase but difficult to increase in backward compatible fashion. Now with introduction of sub_namespace and directly allowing upto 360char keys in LDK, we need support for longer key lengths.

Earlier it was 120 based on uuid length.

### DIFF
--- a/app/src/main/java/org/vss/impl/postgres/sql/v0_create_vss_db.sql
+++ b/app/src/main/java/org/vss/impl/postgres/sql/v0_create_vss_db.sql
@@ -1,6 +1,6 @@
 CREATE TABLE vss_db (
     store_id character varying(120) NOT NULL CHECK (store_id <> ''),
-    key character varying(120) NOT NULL,
+    key character varying(600) NOT NULL,
     value bytea NULL,
     version bigint NOT NULL,
     PRIMARY KEY (store_id, key)


jkczyz:
> Motivation: Earlier we were conservative about our key-size since it is easier to increase but difficult to increase in backward compatible fashion. Now with introduction of sub_namespace and directly allowing upto 360char keys in LDK, we need support for longer key lengths.
> 
> Earlier it was 120 based on uuid length.

What's the math for 360 characters? I'm getting a significantly less number (112), e.g., `monitor_updates/deadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef_4294967295/18446744073709551615`

jkczyz:
Ah, thanks for explaining. Couple questions:
- Can a db's schema be changed after the fact to increase the character limit?
- Any reason why we shouldn't use `text`? (https://www.postgresql.org/docs/current/datatype-character.html)

jkczyz:
Answering one of my own questions, `ALTER COLUMN` can update the type and would cast the data.

https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-DESC-SET-DATA-TYPE

I don't have a strong preference on `text` vs `character varying`, but we should document the choice of the value in the commit message.

jkczyz:
FWIW, I'm seeing opposite advice but it doesn't touch on primary keys, FWIW.

https://wiki.postgresql.org/wiki/Don't_Do_This#Text_storage

Could you update the commit message to say why 600 was chosen?

jkczyz (approved):
Not sure if it makes sense to use a power of two for any efficiency gains, but otherwise LGTM.

PR #28, opened 2024-04-30
Add AuthException

G8XSU:
It will be used when an unauthorized request is made or we fail to authenticate from headers.


app/src/main/java/org/vss/exception/AuthException.java
```diff
@@ -0,0 +1,7 @@
+package org.vss.exception;
+
+public class AuthException extends RuntimeException {
+  public AuthException(String message) {
+    super(message);
+  }
+}
```
jkczyz, line 7: Missing newline character

jkczyz (approved):

jkczyz (approved):

PR #30, opened 2024-07-25
Add created_at and last_updated_at timestamps

G8XSU:
We use application level timestamps to maintain
dialect compatibility between different SQL variations. We truncate timestamp to granularity of days for privacy reasons.


jkczyz:
Change looks good, but I have a couple questions:
- By "dialect compatibility" do you mean implementations use say seconds vs milliseconds, so we need to normalize on something with common granularity?
- What are the new columns used for?

app/src/test/java/org/vss/impl/postgres/PostgresBackendImplIntegrationTest.java
```diff
@@ -48,9 +48,11 @@ void destroy() throws Exception {
   private void createTable(DSLContext dslContext) {
     dslContext.execute("CREATE TABLE vss_db ("
         + "store_id character varying(120) NOT NULL CHECK (store_id <> ''),"
-        + "key character varying(120) NOT NULL,"
+        + "key character varying(600) NOT NULL,"
```
jkczyz, line 51: Could you make a separate commit for this?

jkczyz (approved):

PR #34, opened 2024-09-30
Add Rust based vss-service. 

G8XSU:
Add Rust based vss-service.

* Add Base workspace setup and rustfmt config.
* Add Base protobuf generation for types.
* Add Vss KVStore api.
* Add Authorizer Api with NoopAuthorizer.
* Add empty postgres KvStore implementation.
* Add VssService and base service setup.
* Add KvStoreTestSuite.

### DIFF
--- a/rust/api/.gitignore
+++ b/rust/api/.gitignore
@@ -1,0 +1,3 @@
+Cargo.lock
+target/**
+src/proto/**


rust/api/src/kv_store.rs
```diff
@@ -1,9 +1,12 @@
 use crate::error::VssError;
 use crate::types::{
-	DeleteObjectRequest, DeleteObjectResponse, GetObjectRequest, GetObjectResponse,
+	DeleteObjectRequest, DeleteObjectResponse, GetObjectRequest, GetObjectResponse, KeyValue,
 	ListKeyVersionsRequest, ListKeyVersionsResponse, PutObjectRequest, PutObjectResponse,
 };
 use async_trait::async_trait;
+use bytes::Bytes;
+use rand::distributions::Alphanumeric;
+use rand::{thread_rng, Rng};
```
jkczyz, line 9: I'm seeing compilation errors for this. Could you properly mark anything test-related with `#[cfg(test)]`? Or if this module is entirely going to be for tests, could you mark the entire module accordingly.

rust/api/src/kv_store.rs
```diff
@@ -22,3 +25,602 @@ pub trait KvStore: Send + Sync {
 		&self, user_token: String, request: ListKeyVersionsRequest,
 	) -> Result<ListKeyVersionsResponse, VssError>;
 }
+
+#[macro_export]
+macro_rules! define_kv_store_tests {
```
jkczyz, line 30: Guessing this will be instantiated once Postgres implementation is complete?

rust/api/src/lib.rs
```diff
@@ -1 +1,3 @@
+pub mod error;
```
jkczyz, line 2: Could you add `#![deny(rustdoc::broken_intra_doc_links)]` here and in other crates. Otherwise, there's a lot of broken doc links that are not being caught.

rust/api/src/error.rs
```diff
@@ -0,0 +1,54 @@
...
+}
+
+impl Display for VssError {
+	fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+		match self {
+			VssError::NoSuchKeyError(message) => {
+				write!(f, "Requested key does not exist: {}", message)
+			},
+			VssError::InvalidRequestError(message) => {
+				write!(f, "Request sent to VSS Storage was invalid: {}", message)
```
jkczyz, line 33: Storage? Service? Server?

rust/api/src/error.rs
```diff
@@ -0,0 +1,54 @@
...
+			},
+			VssError::ConflictError(message) => {
+				write!(f, "Potential version conflict in write operation: {}", message)
```
jkczyz, line 36: Potential?

rust/api/src/auth.rs
```diff
@@ -0,0 +1,30 @@
...
+}
+
+// Interface for authorizer that is run before every request.
+#[async_trait]
+pub trait Authorizer: Send + Sync {
+	async fn verify(&self, headers_map: &HashMap<String, String>)
+		-> Result<AuthResponse, VssError>;
+}
+
+// A no-operation authorizer, that lets any user-request go through.
```
jkczyz, line 18: s/that/which

rust/api/src/auth.rs
```diff
@@ -0,0 +1,30 @@
...
+pub struct NoopAuthorizer {}
+
+const UNAUTHENTICATED_USER: &str = "unauth-user";
+
+#[async_trait]
+impl Authorizer for NoopAuthorizer {
+	async fn verify(
+		&self, _headers_map: &HashMap<String, String>,
+	) -> Result<AuthResponse, VssError> {
+		Ok(AuthResponse { user_token: UNAUTHENTICATED_USER.to_string().clone() })
```
jkczyz, line 28: Clone not needed.

rust/api/src/auth.rs
```diff
@@ -0,0 +1,30 @@
...
+
+#[derive(Debug, Clone)]
+pub struct AuthResponse {
+	pub user_token: String,
+}
+
+// Interface for authorizer that is run before every request.
+#[async_trait]
+pub trait Authorizer: Send + Sync {
+	async fn verify(&self, headers_map: &HashMap<String, String>)
```
jkczyz, line 14: This should be documented. Let's add `#![deny(missing_docs)]` to the crates.

rust/server/src/main.rs
```diff
@@ -0,0 +1,75 @@
...
+		);
+		let rest_svc_listener =
+			TcpListener::bind(&addr).await.expect("Failed to bind listening port");
+		loop {
+			tokio::select! {
+				res = rest_svc_listener.accept() => {
+					match res {
+						Ok((stream, _)) => {
+							let io_stream = TokioIo::new(stream);
+							let vss_service = VssService::new(Arc::clone(&store) as Arc<dyn KvStore>,authorizer );
```
jkczyz, line 54: Compilation error here. Need to `Arc::clone(authorizer)`.

rust/server/src/vss_service.rs
```diff
@@ -0,0 +1,154 @@
...
+	PutObjectResponse,
+};
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+
+#[derive(Clone)]
+pub struct VssService {
+	store: Arc<dyn KvStore>,
+	authorizer: Arc<dyn Authorizer>,
```
jkczyz, line 24: Why not use type parameters?

rust/server/src/vss_service.rs
```diff
@@ -0,0 +1,154 @@
...
+>(
+	store: Arc<dyn KvStore>, authorizer: Arc<dyn Authorizer>, request: Request<Incoming>,
+	handler: F,
+) -> Result<<VssService as Service<Request<Incoming>>>::Response, hyper::Error> {
+	// TODO: we should bound the amount of data we read to avoid allocating too much memory.
+	let (parts, body) = request.into_parts();
+	let headers_map = parts
+		.headers
+		.iter()
+		.map(|(k, v)| (k.as_str().to_string(), v.to_str().unwrap_or("").to_string()))
```
jkczyz, line 102: Should we consider passing the `HeaderMap` instead to avoid these allocations and conversion of non-ASCII values to something unusable?

rust/server/src/vss_service.rs
```diff
@@ -0,0 +1,154 @@
...
+async fn handle_request<
+	T: Message + Default,
+	R: Message,
+	F: FnOnce(Arc<dyn KvStore>, String, T) -> Fut + Send + 'static,
+	Fut: Future<Output = Result<R, VssError>> + Send,
+>(
+	store: Arc<dyn KvStore>, authorizer: Arc<dyn Authorizer>, request: Request<Incoming>,
+	handler: F,
+) -> Result<<VssService as Service<Request<Incoming>>>::Response, hyper::Error> {
+	// TODO: we should bound the amount of data we read to avoid allocating too much memory.
```
jkczyz, line 97: This comment should be on `let bytes = body.collect().await?.to_bytes();`, IIUC.

rust/server/src/vss_service.rs
```diff
@@ -0,0 +1,154 @@
...
+		},
+		Err(_) => Ok(Response::builder()
+			.status(StatusCode::BAD_REQUEST)
+			.body(Full::new(Bytes::from(b"Error parsing request".to_vec())))
+			// unwrap safety: body only errors when previous chained calls failed.
+			.unwrap()),
+	}
+}
+
+fn build_error_response(e: VssError) -> Result<Response<Full<Bytes>>, Error> {
```
jkczyz, line 127: Doesn't look like this needs to return a `Result`.

jkczyz (commented):
Concept ACK. Need to read through the tests still.

rust/server/src/vss_service.rs
jkczyz: Sure, in `main.rs` it could be parameterized that way, but that doesn't require having `dyn` leak into the library.

jkczyz: There's a runtime cost for `dyn` as it results in dynamic dispatch and prevents the compiler from inlining. The tradeoff is larger binary size as the code will be duplicated for each instantiation.

https://doc.rust-lang.org/std/keyword.dyn.html#trade-offs

rust/api/src/kv_store.rs
```diff
@@ -22,3 +25,602 @@ pub trait KvStore: Send + Sync {
...
+			let current_page = if previous_page.is_none() {
+				let page = ctx.list(None, None, None).await?;
+				assert_eq!(page.global_version, Some(1003));
+				page
+			} else {
+				let next_page_token = previous_page.as_ref().unwrap().next_page_token.clone();
+				let page = ctx.list(next_page_token, None, None).await?;
+				assert!(page.global_version.is_none());
+				page
+			};
```
jkczyz, line 422: nit: can `match` on `previous_page.take()` to avoid `unwrap` and `clone`

rust/api/src/kv_store.rs
```diff
@@ -22,3 +25,602 @@ pub trait KvStore: Send + Sync {
...
+			if current_page.key_versions.is_empty() {
+				break;
+			}
+
+			all_key_versions.extend(current_page.key_versions.clone());
+			previous_page = Some(current_page);
+		}
+
+		let unique_keys: std::collections::HashSet<String> =
+			all_key_versions.iter().map(|kv| kv.key.clone()).collect();
```
jkczyz, line 433: nit: use `into_iter` to avoid cloning

rust/api/src/kv_store.rs
```diff
@@ -22,3 +25,602 @@ pub trait KvStore: Send + Sync {
...
+				let page = ctx.list(next_page_token, None, None).await?;
+				assert!(page.global_version.is_none());
+				page
+			};
+
+			if current_page.key_versions.is_empty() {
+				break;
+			}
+
+			all_key_versions.extend(current_page.key_versions.clone());
```
jkczyz, line 428: nit: can avoid the clone by storing just the `next_page_token` instead of the entire `ListKeyVersionsResponse`

rust/api/src/kv_store.rs
```diff
@@ -22,3 +25,602 @@ pub trait KvStore: Send + Sync {
...
+		if let Some(k2_response) = all_key_versions.iter().find(|kv| kv.key == "k2") {
+			assert_eq!(k2_response.key, "k2");
+			assert_eq!(k2_response.version, 3);
+			assert_eq!(k2_response.value, Bytes::new());
+		}
+
+		Ok(())
+	}
+
+	async fn list_should_honour_page_size_and_key_prefix_if_provided() -> Result<(), VssError> {
```
jkczyz, line 452: Likewise for these tests.

rust/api/src/kv_store.rs
```diff
@@ -22,3 +25,602 @@ pub trait KvStore: Send + Sync {
...
+
+#[async_trait]
+pub(crate) trait KvStoreTestSuite {
+	type Store: KvStore + 'static;
+
+	async fn create_store() -> Self::Store;
+
+	async fn put_should_succeed_when_single_object_put_operation() -> Result<(), VssError> {
+		let kv_store = Self::create_store().await;
+		let ctx = TestContext::new(&kv_store);
```
jkczyz, line 101: Can DRY these tests up by having each take a `&TestContext` created in `create_test!`.

jkczyz (commented):
Mostly nits on the tests. Verified that they match the Java integration tests. https://github.com/lightningdevkit/vss-server/blob/main/java/app/src/test/java/org/vss/AbstractKVStoreIntegrationTest.java

rust/api/src/kv_store.rs
```diff
@@ -74,6 +77,7 @@ macro_rules! define_kv_store_tests {
 }
 
 #[async_trait]
+#[cfg(test)]
```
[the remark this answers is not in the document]
jkczyz: Ah, I think we typically use a `_test_utils` feature for something like this.

https://github.com/lightningdevkit/rust-lightning/blob/8da30df223d50099c75ba8251615bd2026fcea75/lightning/Cargo.toml#L20C1-L20C12

rust/server/src/vss_service.rs
jkczyz: Hmm... I guess in LDK Node we are supporting that but just hiding `dyn` in a type alias.

https://github.com/lightningdevkit/ldk-node/blob/aaccde78ce072d2eed345c7c0cc23006eb0d4d2a/src/types.rs#L34

So maybe fine to just do the same.

rust/api/src/kv_store_tests.rs
```diff
@@ -0,0 +1,612 @@
...
+			key_prefix,
+		};
+		let response = self.kv_store.list_key_versions(self.user_token.clone(), request).await?;
+		Ok(response)
+	}
+}
+
+fn kv(key: &str, value: &str, version: i64) -> KeyValue {
+	KeyValue { key: key.to_string(), version, value: Bytes::from(value.to_string()) }
+}
```
jkczyz, line 612: Missing newline

jkczyz (approved):
LGTM. Let's squash and merge.

jkczyz (approved):

PR #35, opened 2024-09-30
Add protocol compliant PostgresKvStore impl.

G8XSU:
* Add protocol compliant PostgresKvStore impl.

### DIFF
--- a/rust/impls/src/postgres_store.rs
+++ b/rust/impls/src/postgres_store.rs
@@ -1,19 +1,207 @@
 use api::error::VssError;
-use api::kv_store::KvStore;
+use api::kv_store::{KvStore, GLOBAL_VERSION_KEY, INITIAL_RECORD_VERSION};
 use api::types::{
-	DeleteObjectRequest, DeleteObjectResponse, GetObjectRequest, GetObjectResponse,
+	DeleteObjectRequest, DeleteObjectResponse, GetObjectRequest, GetObjectResponse, KeyValue,
 	ListKeyVersionsRequest, ListKeyVersionsResponse, PutObjectRequest, PutObjectResponse,
 };
 use async_trait::async_trait;
-use std::io::Error;
+use bb8_postgres::bb8::Pool;
+use bb8_postgres::PostgresConnectionManager;
+use bytes::Bytes;
+use chrono::Utc;
+use std::cmp::min;
+use std::io;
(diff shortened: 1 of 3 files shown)


rust/impls/src/postgres_store.rs
```diff
@@ -227,28 +239,30 @@ impl KvStore for PostgresBackendImpl {
...
+			.map(|kv| self.build_vss_record(user_token.to_string(), store_id.to_string(), kv))
 			.collect();
 
 		if let Some(global_version) = request.global_version {
 			let global_version_record = self.build_vss_record(
-				&user_token,
-				store_id,
-				&KeyValue {
+				user_token,
+				store_id.to_string(),
```
jkczyz, line 264: `to_string()` not needed here.

rust/impls/src/postgres_store.rs
```diff
@@ -300,7 +314,10 @@ impl KvStore for PostgresBackendImpl {
 		&self, user_token: String, request: DeleteObjectRequest,
 	) -> Result<DeleteObjectResponse, VssError> {
 		let store_id = &request.store_id;
-		let vss_record = self.build_vss_record(&user_token, store_id, &request.key_value.unwrap());
+		let key_value = request.key_value.ok_or_else(|| {
+			VssError::InvalidRequestError("key_value missing in DeleteObjectRequest".to_string())
+		})?;
+		let vss_record = self.build_vss_record(user_token, store_id.to_string(), key_value);
```
jkczyz, line 320: `to_string()` not needed here.

jkczyz (approved):
Just two minor comments, otherwise LGTM after squash.

rust/impls/src/postgres_store.rs
```diff
@@ -227,28 +239,30 @@ impl KvStore for PostgresBackendImpl {
 		&self, user_token: String, request: PutObjectRequest,
 	) -> Result<PutObjectResponse, VssError> {
 		let store_id = &request.store_id;
```
jkczyz, line 241: No need to take a reference here. Without it you move the value, and then calling `to_string` later is equivalent to `clone`. Then you don't need it for the last use because it is moved. Likewise in `delete`. Removing the references would allow addressing the other two comments.

jkczyz (approved):

PR #46, opened 2025-03-14
Add file-based config support.

G8XSU:
* Add file-based config support.

### DIFF
--- a/rust/server/src/main.rs
+++ b/rust/server/src/main.rs
@@ -23,11 +23,32 @@ use api::kv_store::KvStore;
 use impls::postgres_store::PostgresBackendImpl;
 use std::sync::Arc;
 
+pub(crate) mod util;
 pub(crate) mod vss_service;
 
 fn main() {
-	// Define the address to bind the server to
-	let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
+	let args: Vec<String> = std::env::args().collect();
+	if args.len() != 2 {
+		eprintln!("Usage: {} <config-file-path>", args[0]);
+		std::process::exit(1);
+	}
+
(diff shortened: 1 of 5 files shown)


rust/server/src/util/config.rs
```diff
@@ -0,0 +1,46 @@
...
+impl PostgreSQLConfig {
+	pub(crate) fn to_connection_string(&self) -> String {
+		let username = std::env::var("VSS_POSTGRESQL_USERNAME")
+			.ok()
+			.or_else(|| self.username.clone())
+			.expect("PostgreSQL database username must be provided in config or env var VSS_POSTGRESQL_USERNAME must be set.");
+		let password = std::env::var("VSS_POSTGRESQL_PASSWORD")
+			.ok()
+			.or_else(|| self.password.clone())
+			.expect("PostgreSQL database password must be provided in config or env var VSS_POSTGRESQL_PASSWORD must be set.");
```
jkczyz, line 33: nit: Might be a way to avoid these clones by using `as_ref` on the `Option`s.

rust/server/src/main.rs
```diff
@@ -23,11 +23,32 @@ use api::kv_store::KvStore;
 use impls::postgres_store::PostgresBackendImpl;
 use std::sync::Arc;
 
+pub(crate) mod util;
 pub(crate) mod vss_service;
 
 fn main() {
-	// Define the address to bind the server to
-	let addr = SocketAddr::from(([127, 0, 0, 1], 3000));
+	let args: Vec<String> = std::env::args().collect();
```
jkczyz, line 30: nit: Instead of collecting we can simply use the `Iterator` interface implemented by `Args`.

jkczyz (approved):

PR #50, opened 2025-07-03
build: upgrade jooq and add junit platform

enigbe:
## What this PR does
With the release of Gradle `9.0.0`, we experienced build
failures due to deprecated access to plugin conventions
(JavaPluginConvention).
Ref: [Gradle `9.0` release]( https://docs.gradle.org/9.0.0-rc-1/userguide/upgrading_version_8.html#deprecated_access_to_conventions)

To address this, we:
- Upgrade jOOQ Gradle plugin from `8.0` to `8.2.2` to
avoid deprecation warnings when using Gradle `9.0`.
Ref: [jOOQ plugin release](https://github.com/etiennestuder/gradle-jooq-plugin/releases/tag/v8.2.2)
- Upgrade jOOQ from `3.17.7` to `3.19.1` to match the
plugin version upgrade.

Note: We avoid the latest version of jOOQ because it
requires Java 21.

Additionally, we add JUnit Platform Launcher which
is required for JUnit 5 tests in newer Gradle versions.

## Related Issues
- LDK-Node Issue [#581](https://github.com/lightningdevkit/ldk-node/issues/581)

cc @tnull

### DIFF
--- a/.github/workflows/build-and-deploy-java.yml
+++ b/.github/workflows/build-and-deploy-java.yml
@@ -54,10 +54,15 @@ jobs:
         run: |
           # Print Info
           java -version
-          gradle --version
+          GRADLE_VERSION=$(gradle --version | awk '/^Gradle/ {print $2}' | head -1)
+          if [ -z "$GRADLE_VERSION" ]; then
+            echo "Failed to extract Gradle version, using default"
+            GRADLE_VERSION="9.0"
+          fi
+          echo "Gradle $GRADLE_VERSION"
 
           cd java
-          gradle wrapper --gradle-version 8.1.1
+          gradle wrapper --gradle-version $GRADLE_VERSION
           ./gradlew --version
           ./gradlew build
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-deploy-java.yml
```diff
@@ -54,10 +54,15 @@ jobs:
         run: |
           # Print Info
           java -version
-          gradle --version
+          GRADLE_VERSION=$(gradle --version | awk '/^Gradle/ {print $2}' | head -1)
```
enigbe, line 57: Instead of using hard-coded Gradle versions, we extract the version number from the current release.

enigbe:
> The changes make sense to me, but AFAICT the CI over at https://github.com/enigbe/ldk-node/pull/5 still fails due to expecting Gradle 10? Do we need to fix this here or over on the LDK Node side?

As is, it is meant to fail until these changes go in.
Here is the [passing CI test](https://github.com/enigbe/ldk-node/actions/runs/16121950046) after introducing commit [430275a](https://github.com/enigbe/ldk-node/pull/5/commits/430275a5588c0e6915a7db481e495f295e8e61bc) that checks out these changes from my fork.

PR #59, opened 2025-10-12
feat: Added in-memory storage for testing purposes

Harshdev098:
Have added in_memory store for testing purpose.
We can edit config file to use specific store either postgresql or memory


tnull (commented):
Thanks for looking into this!

Generally goes into the right direction, but we def. need to avoid re-allocating everything on every operation.

tnull (commented):
Looks much better, but I think we still need to handle `global_version` properly, even if we're currently not using it client-side.

tnull (commented):
One comment, will take another look once @tankyleo also had a chance to do a review round here.

Harshdev098:
@tankyleo Can you please review it!

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,367 @@
...
+			store_id,
+			key: kv.key,
+			value: kv.value.to_vec(),
+			version: kv.version,
+			created_at: now,
+			last_updated_at: now,
+		}
+	}
+
+	fn build_key(user_token: &str, store_id: &str, key: &str) -> String {
```
tankyleo: Same here pull this out into a helper function in this file

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,367 @@
...
+	store: &mut HashMap<String, VssDbRecord>, record: VssDbRecord,
+) -> Result<(), VssError> {
+	let key = format!("{}#{}#{}", record.user_token, record.store_id, record.key);
+	let now = Utc::now();
+
+	let entry = store.entry(key);
+	match entry {
+		std::collections::hash_map::Entry::Occupied(mut occ) => {
+			let existing = occ.get_mut();
+			if existing.version >= record.version {
```
tankyleo: For already existing records, a conflict error should be returned anytime the versions don't match, unless `record.version` has been set to `-1` (meaning non-conditional update).

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,367 @@
...
+}
+
+fn execute_delete_object(
+	store: &mut HashMap<String, VssDbRecord>, record: VssDbRecord,
+) -> Result<(), VssError> {
+	let key = format!("{}#{}#{}", record.user_token, record.store_id, record.key);
+	if let Some(existing) = store.get(&key) {
+		if existing.version != record.version {
+			return Ok(());
+		}
```
tankyleo: If `record.version` is -1, this should be skipped and we should do a non-conditional delete.

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,367 @@
...
+
+		if let Some(record) = guard.get(&key) {
+			Ok(GetObjectResponse {
+				value: Some(KeyValue {
+					key: record.key.clone(),
+					value: Bytes::from(record.value.clone()),
+					version: record.version,
+				}),
+			})
+		} else if request.key == GLOBAL_VERSION_KEY {
```
tankyleo: Looks like by the time we are here, we know the `GLOBAL_VERSION_KEY` does not have a value, otherwise `guard.get` would have returned `Some` previously. We can just return the `GetObjectResponse` below directly with `version: 0`.

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,367 @@
...
+
+		Ok(DeleteObjectResponse {})
+	}
+
+	async fn list_key_versions(
+		&self, user_token: String, request: ListKeyVersionsRequest,
+	) -> Result<ListKeyVersionsResponse, VssError> {
+		let store_id = request.store_id;
+		let key_prefix = request.key_prefix.unwrap_or_default();
+		let page_token = request.page_token.unwrap_or_default();
```
tankyleo: If you don't mind, let's use the more explicit `unwrap_or` here like we have just below. `unwrap_or_default` is a little hard to read.

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,367 @@
...
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use super::*;
+	use api::kv_store::INITIAL_RECORD_VERSION;
+	use bytes::Bytes;
+	use tokio::test;
+
```
tankyleo: Like we have in `postgres_store`, let's add a line similar to below here and make sure the tests pass
```
define_kv_store_tests!(InMemoryKvStoreTest, InMemoryBackendImpl, InMemoryBackendImpl::new());
```

tankyleo (commented):
Sorry for the delay !

Harshdev098:
@tankyleo Have done with the required changes! Can you please review it

tnull (commented):
When testing integration with LDK Node locally I found that the tests are currently failing. I now opened https://github.com/lightningdevkit/vss-server/pull/62 to add LDK Node integration tests to our CI here. It would be great if that could land first, and we could also add a CI job for the in-memory store as part of this PR then, ensuring the implementation actually works as expected.

Harshdev098:
Hey @tnull I have updated the PR the checks are passing, will add a CI job for it and update you!

rust/api/src/kv_store_tests.rs
```diff
@@ -428,34 +428,33 @@ pub trait KvStoreTestSuite {
 		let total_kv_objects = 20;
 		let page_size = 5;
 		for i in 0..total_kv_objects {
-			ctx.put_objects(Some(i as i64), vec![kv(&format!("{}k", i), "k1v1", 0)]).await?;
+			ctx.put_objects(None, vec![kv(&format!("{}k", i), "k1v1", 0)]).await?;
```
tnull: Can you explain why you need to change any of the test cases at all? Shouldn't they all just pass when run against the in-memory store?

tnull (commented):
> Hey @tnull @tankyleo I have added the CI Job for the tests and also have updated the main function to accept `in-memory` flag to override the config file. Can you please review it!

Thanks, added some comments. Please also make sure that the LDK Node CI is run against the in-memory store.

tnull (commented):
> Please also make sure that the LDK Node CI is run against the in-memory store.

AFAICT this is still missing the part that runs the LDK Node integration test against the in-memory store to check  functionality in a realistic setting?

tnull (commented):
Right, and now you're hitting the errors that I originally found when running LDK Node against the in-memory store. Very likely, they are due to the in-memory implementation behaving in an unexpected manner. We need to fix this before this PR can get merged. Please take a look, but let me know if you end up getting stuck or need some assistance figuring this out.

Harshdev098:
Hey @tnull @tankyleo Have tested it locally, the test are working correctly now, don't know why its stuck in CI! Can you please review it

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,463 @@
...
+		let store_id = request.store_id.clone();
+		let mut guard = self.store.write().await;
+		let vss_record = build_vss_record(user_token.clone(), store_id.clone(), key_value);
+
+		execute_delete_object(&mut guard, &user_token, &store_id, &vss_record);
+
+		Ok(DeleteObjectResponse {})
+	}
+
+	async fn list_key_versions(
```
Harshdev098: There was a deadlock here earlier have resolved that! now its running locally

rust/impls/src/in_memory_store.rs
tankyleo: @Harshdev098 help me understand we still have the `unwrap_or_default` here ? If you would prefer `unwrap_or_default` let me know.

rust/impls/src/postgres_store.rs
```diff
@@ -93,10 +93,7 @@ async fn drop_database(postgres_endpoint: &str, db_name: &str) -> Result<(), Err
 
 	let drop_database_statement = format!("{} {};", DROP_DB_CMD, db_name);
 	let num_rows = client.execute(&drop_database_statement, &[]).await.map_err(|e| {
-		Error::new(
-			ErrorKind::Other,
-			format!("Failed to drop database {}: {}", db_name, e),
-		)
+		Error::new(ErrorKind::Other, format!("Failed to drop database {}: {}", db_name, e))
```
tankyleo: Sorry the formats in this file were done in a CI-specific PR we merged, go ahead and rebase and drop them thank you :)

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,463 @@
...
+	VssDbRecord {
+		user_token,
+		store_id,
+		key: kv.key,
+		value: kv.value.to_vec(),
+		version: kv.version,
+		created_at: now,
+		last_updated_at: now,
+	}
+}
```
tankyleo: Once we only build the `VssDbRecord` when we are about to insert into the database, we can delete this function here.

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,463 @@
...
+				last_updated_at: now,
+			};
+			vac.insert(new_record);
+		},
+	}
+}
+
+fn execute_conditional_insert(
+	store: &mut HashMap<String, VssDbRecord>, user_token: &str, store_id: &str,
+	record: &VssDbRecord,
```
tankyleo: Same reasoning as above, let's pass a `key_value: KeyValue` here

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,463 @@
...
+		version: INITIAL_RECORD_VERSION as i64,
+		created_at: now,
+		last_updated_at: now,
+	};
+	store.insert(key, new_record);
+}
+
+fn execute_conditional_update(
+	store: &mut HashMap<String, VssDbRecord>, user_token: &str, store_id: &str,
+	record: &VssDbRecord,
```
tankyleo: Same here, a `key_value: KeyValue` fits better since we are only modifying an existing `VssDbRecord`, not creating a new one.

tankyleo (commented):
Sorry for the delay again I have cleared all the other priorities, this is now top priority :) Here are some comments, mostly on the types we pass to the different functions.

Will continue review tomorrow 100% ! Thank you again.

rust/impls/src/postgres_store.rs
Harshdev098: Ohh sorry @tankyleo , but I have done it already, my branch is now up to date with main

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,414 @@
...
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+fn build_storage_key(user_token: &str, store_id: &str, key: &str) -> String {
+	format!("{}#{}#{}", user_token, store_id, key)
+}
+
+/// In-memory implementation of the VSS Store.
+pub struct InMemoryBackendImpl {
+	store: Arc<RwLock<HashMap<String, VssDbRecord>>>,
```
tankyleo: How about we use a `BTreeMap` here to avoid having to resort all the keys on every subsequent call to `list_key_versions` further down below ?

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,414 @@
...
+
+		if let Some(record) = guard.get(&key) {
+			Ok(GetObjectResponse {
+				value: Some(KeyValue {
+					key: record.key.clone(),
+					value: Bytes::from(record.value.clone()),
+					version: record.version,
+				}),
+			})
+		} else if request.key == GLOBAL_VERSION_KEY {
```
tankyleo: This  spot here *would* benefit from a comment explaining that in all cases were global version is not 0, we would have matched on the previous branch.

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,414 @@
...
+	store: &mut HashMap<String, VssDbRecord>, user_token: &str, store_id: &str, key_value: KeyValue,
+) {
+	if key_value.version == -1 {
+		execute_non_conditional_upsert(store, user_token, store_id, key_value);
+	} else if key_value.version == 0 {
+		execute_conditional_insert(store, user_token, store_id, key_value);
+	} else {
+		execute_conditional_update(store, user_token, store_id, key_value);
+	}
+}
```
tankyleo: We can delete all the branches and intermediate functions here and have a single function that looks like this below. We've already done all the validation, so here we can just focus on inserting / updating the value.

```rust
fn execute_put_object(
        store: &mut HashMap<String, VssDbRecord>, user_token: &str, store_id: &str, key_value: KeyValue,
) {
        let key = build_storage_key(user_token, store_id, &key_value.key);
        let now = Utc::now();

        match store.entry(key) {
                std::collections::hash_map::Entry::Occupied(mut occ) => {
                        let existing = occ.get_mut();
                        existing.version = if key_value.version == -1 {
                                INITIAL_RECORD_VERSION as i64
                        } else {
                                existing.version.saturating_add(1)
                        };
                        existing.value = key_value.value.to_vec();
                        existing.last_updated_at = now;
                },
                std::collections::hash_map::Entry::Vacant(vac) => {
                        let new_record = VssDbRecord {
                                user_token: user_token.to_string(),
                                store_id: store_id.to_string(),
                                key: key_value.key,
                                value: key_value.value.to_vec(),
                                version: INITIAL_RECORD_VERSION as i64,
                                created_at: now,
                                last_updated_at: now,
                        };
                        vac.insert(new_record);
                },
        }
}
```

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,414 @@
...
+			let guard = self.store.read().await;
+
+			let mut global_version: Option<i64> = None;
+			if page_token_option.is_none() {
+				global_version =
+					Some(self.get_current_global_version(&guard, &user_token, &store_id));
+			}
+
+			let storage_prefix = format!("{}#{}#", user_token, store_id);
+			let mut temp: Vec<(String, i64)> = Vec::new();
```
tankyleo: Let's try to remove this allocation here :) If we use a `BTreeMap` instead as described earlier, I think we can do everything we need in a single iterator chain using things like `filter`, `skip_while`, and `map`

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,414 @@
...
+				value: Bytes::new(),
+				version: *version,
+			})
+			.collect();
+
+		let next_page_token = if page_items.is_empty() {
+			Some("".to_string())
+		} else {
+			page_items.last().map(|kv| kv.key.clone())
+		};
```
tankyleo: How about we use the number of elements we've returned already as the `next_page_token` ? Like this we can just call `skip(next_page_token)` to get straight to our starting point without having to compare the keys with our page_token.

rust/server/src/main.rs
```diff
@@ -28,19 +29,28 @@ pub(crate) mod vss_service;
...
 		Ok(cfg) => cfg,
 		Err(e) => {
 			eprintln!("Failed to load configuration: {}", e);
 			std::process::exit(1);
 		},
 	};
 
+	// Override the `store_type` if --in-memory flag passed
+	if use_in_memory {
+		println!("Overriding backend type: using in-memory backend (via --in-memory flag)");
```
tankyleo: This `println` is not necessary we can remove it thank you

tankyleo (commented):
Some more comments focusing on `in_memory_store` thank you

rust/impls/src/postgres_store.rs
tankyleo: @Harshdev098 want to make sure you know about `git reflog` ? Would have allowed you to revert it back easily before pushing but all good now !

rust/impls/src/in_memory_store.rs
```diff
@@ -288,66 +288,177 @@ impl KvStore for InMemoryBackendImpl {
...
+		Command::new(&target_dir)
+			.arg("vss-server-config.toml")
+			.arg("--in-memory")
+			.current_dir("../server")
+			.spawn()
+			.expect("failed to start server")
+	}
+
+	#[tokio::test]
+	async fn test_put_and_get_via_http() {
```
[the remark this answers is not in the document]
tankyleo: @Harshdev098 where did this test go ? I don't see it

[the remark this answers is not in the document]
tankyleo: Ah seems like we removed it ok.

rust/server/src/main.rs
```diff
@@ -28,19 +29,28 @@ pub(crate) mod vss_service;
 
 fn main() {
 	let args: Vec<String> = std::env::args().collect();
-	if args.len() != 2 {
-		eprintln!("Usage: {} <config-file-path>", args[0]);
+	if args.len() < 2 {
+		eprintln!("Usage: {} <config-file-path> [--in-memory]", args[0]);
```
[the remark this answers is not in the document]
Harshdev098: I added `--in-memory` to make it easy for external projects like Fedimint to run integration tests with our VSS server

rust/impls/src/in_memory_store.rs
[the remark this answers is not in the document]
Harshdev098: Switching back to Mutex, it was deadlock issue at that time!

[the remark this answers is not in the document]
Harshdev098: Also I found that testing the ldk node integration test against memory_store I need to restart the server to test after every updation, because I think starting the integration test it needs fresh data but it gives error when runned multiple times without restarting the server, which can make bad experience as a developer @tankyleo!?

rust/server/src/main.rs
```diff
@@ -28,19 +29,28 @@ pub(crate) mod vss_service;
...
 		Err(e) => {
 			eprintln!("Failed to load configuration: {}", e);
 			std::process::exit(1);
 		},
 	};
 
+	// Override the `store_type` if --in-memory flag passed
+	if use_in_memory {
+		println!("Overriding backend type: using in-memory backend (via --in-memory flag)");
+		config.server_config.store_type = "in_memory".to_string();
```
enigbe: This would need to be modified if you accept the `StoreType` enum recommendation.

rust/server/src/util/config.rs
```diff
@@ -10,6 +10,7 @@ pub(crate) struct Config {
 pub(crate) struct ServerConfig {
 	pub(crate) host: String,
 	pub(crate) port: u16,
+	pub(crate) store_type: String, // "postgresql" or "in_memory"
```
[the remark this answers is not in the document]
Harshdev098: Ofcourse, better way to handle that! Thanks.

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,361 @@
...
+					key: GLOBAL_VERSION_KEY.to_string(),
+					value: Bytes::new(),
+					version: 0,
+				}),
+			})
+		} else {
+			Err(VssError::NoSuchKeyError("Requested key not found.".to_string()))
+		};
+
+		result
```
tankyleo: nit: no need for the extra result variable here, just return from the if statement directly

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,361 @@
...
+
+				Some(KeyValue {
+					key: key.to_string(),
+					value: Bytes::new(),
+					version: record.version,
+				})
+			})
+			.collect();
+
+		all_items.sort_by(|a, b| a.key.cmp(&b.key));
```
tankyleo: Again as I mentioned further above, we now don't need this `sort` here let's delete it

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,361 @@
...
+
+		let mut all_items: Vec<KeyValue> = guard
+			.iter()
+			.filter(|(storage_key, _)| storage_key.starts_with(&storage_prefix))
+			.filter_map(|(storage_key, record)| {
+				let key = &storage_key[prefix_len..];
+
+				if key == GLOBAL_VERSION_KEY {
+					return None;
+				}
```
tankyleo: Delete this once we filter out the global version key in the first filter call.

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,361 @@
...
+			global_version = Some(self.get_current_global_version(&guard, &user_token, &store_id));
+		}
+
+		let storage_prefix = format!("{}#{}#", user_token, store_id);
+		let prefix_len = storage_prefix.len();
+
+		let mut all_items: Vec<KeyValue> = guard
+			.iter()
+			.filter(|(storage_key, _)| storage_key.starts_with(&storage_prefix))
+			.filter_map(|(storage_key, record)| {
```
tankyleo: Let's not do any filtering in this function, just `map`

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,361 @@
...
+					value: Bytes::new(),
+					version: record.version,
+				})
+			})
+			.collect();
+
+		all_items.sort_by(|a, b| a.key.cmp(&b.key));
+
+		let page_items: Vec<KeyValue> =
+			all_items.iter().skip(offset).take(limit).cloned().collect();
```
tankyleo: As mentioned earlier, this `skip(offset).take(limit)` can be folded into the first iterator chain.

rust/server/vss-server-config.toml
```diff
@@ -1,6 +1,7 @@
 [server_config]
 host = "127.0.0.1"
 port = 8080
+store_type = "postgres" # "postgres" for using postgresql and "in_memory" for testing purposes
```
tankyleo: `in_memory` is not valid any longer

tankyleo (commented):
Can you also rephrase your commit messages to use the imperative form ? See https://cbea.ms/git-commit/

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,361 @@
...
+	async fn list_key_versions(
+		&self, user_token: String, request: ListKeyVersionsRequest,
+	) -> Result<ListKeyVersionsResponse, VssError> {
+		let store_id = request.store_id.clone();
+		let key_prefix = request.key_prefix.clone().unwrap_or_default();
+		let page_size = request.page_size.unwrap_or(i32::MAX);
+		let limit = std::cmp::min(page_size, LIST_KEY_VERSIONS_MAX_PAGE_SIZE) as usize;
+
+		let offset: usize =
+			request.page_token.as_ref().and_then(|s| s.parse::<usize>().ok()).unwrap_or(0);
```
tankyleo: If we cannot parse the token, that is an error - currently we accept it and set `offset` to 0.

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,350 @@
...
+use std::sync::Arc;
+use tokio::sync::Mutex;
+
+fn build_storage_key(user_token: &str, store_id: &str, key: &str) -> String {
+	format!("{}#{}#{}", user_token, store_id, key)
+}
+
+/// In-memory implementation of the VSS Store.
+pub struct InMemoryBackendImpl {
+	store: Arc<Mutex<BTreeMap<String, VssDbRecord>>>,
```
tnull: No need to hold this in an `Arc` AFAICT.
```suggestion
	store: Mutex<BTreeMap<String, VssDbRecord>>,
```

.github/workflows/ldk-node-integration.yml
```diff
@@ -30,20 +29,105 @@ jobs:
...
         with:
           path: vss-server
+
       - name: Checkout LDK Node
         uses: actions/checkout@v3
         with:
           repository: lightningdevkit/ldk-node
           path: ldk-node
 
-      - name: Build and Deploy VSS Server
```
tnull: While these ~unrelated changes here and elsewhere look fine, I want to note that it would have been preferable if they happened in separate (prefactor) commits. It would at the very least have made reviewing a lot easier.

rust/impls/src/in_memory_store.rs
tankyleo: The first lookup handles cases where the `GLOBAL_VERSION_KEY` is some non-zero value. We want to check whether it's been set to some value in the map before returning the initial value in this branch.

tnull (commented):
This needs a rebase now that https://github.com/lightningdevkit/vss-server/pull/67 landed.

rust/server/Cargo.toml
```diff
@@ -21,3 +21,7 @@ prost = { version = "0.11.6", default-features = false, features = ["std"] }
 bytes = "1.4.0"
 serde = { version = "1.0.203", default-features = false, features = ["derive"] }
 toml = { version = "0.8.9", default-features = false, features = ["parse"] }
+
+[[bin]]
+name = "vss-server"
+path = "src/main.rs"
```
tankyleo: Help me understand I don't think we need this ?

tankyleo (commented):
Thanks for the rebase !

Can you squash the 3rd commit into the 1st commit ? Let's ship the improved version in a single commit.

Also CI is failing can you take a look ? Thank you

Harshdev098:
@tankyleo Have squashed those commits

rust/server/Cargo.toml
tankyleo: @Harshdev098 before I take a deeper look can you explain this one ? It doesn't seem necessary.

Harshdev098: Actually earlier the CI test are failing for not finding a bin or something like that, but now have reverted does changes!

Harshdev098:
@tankyleo Can you please tabke a look at it!

Harshdev098:
@tnull @tankyleo I have rebased it, can you please review it in your mean time!

tnull:
Thanks! Please feel free to re-request review once @tankyleo is happy with the current state.

.github/workflows/build-and-test-in-memory.yml
```diff
@@ -23,25 +23,24 @@ jobs:
           mkdir -p rust/server
           cat > rust/server/vss-server-config.toml <<EOF
           [server_config]
-          host = "127.0.0.1"
-          port = 8080
+          bind_address = "127.0.0.1:8080"
```
tankyleo, line 26: Please make all the changes in this file in a single commit.

.github/workflows/ldk-node-integration.yml
```diff
@@ -35,21 +30,53 @@ jobs:
...
+      - name: Create Postgres config
```
tankyleo, line 40: Let's not make any edits to the CI of the postgres backend in this PR. We'll use a `cfg` flag to enable the in-memory backend so that no edits to the config file will be needed.

.github/workflows/ldk-node-integration.yml
```diff
@@ -74,16 +101,15 @@ jobs:
           mkdir -p vss-server/rust/server
           cat > vss-server/rust/server/vss-server-config.toml <<EOF
           [server_config]
-          host = "127.0.0.1"
-          port = 8080
+          bind_address = "127.0.0.1:8080"
```
tankyleo, line 104: Please keep all your changes for the CI of the in-memory server to a single commit.

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,347 @@
...
+		}
+	}
+}
+
+fn validate_delete_operation(
+	store: &BTreeMap<String, VssDbRecord>, user_token: &str, store_id: &str, key_value: &KeyValue,
+) -> Result<(), VssError> {
+	let key = build_storage_key(user_token, store_id, &key_value.key);
+
+	if key_value.version == -1 {
```
tankyleo, line 77: This is used to validate a `PutObjectRequest`, and therefore *should* return a `ConflictError` in case the key does not exist, even in the case of a non-conditional write.

rust/impls/src/in_memory_store.rs
```diff
@@ -0,0 +1,347 @@
...
+		&self, user_token: String, request: DeleteObjectRequest,
+	) -> Result<DeleteObjectResponse, VssError> {
+		let key_value = request.key_value.ok_or_else(|| {
+			VssError::InvalidRequestError("key_value missing in DeleteObjectRequest".to_string())
+		})?;
+
+		let store_id = request.store_id;
+		let mut guard = self.store.lock().await;
+
+		execute_delete_object(&mut guard, &user_token, &store_id, &key_value);
```
tankyleo, line 234: Before we reach this point for conditional deletes, we still need to make sure that the version matches in case the key *does* exist. See `rust/api/src/types.rs` for further details.

rust/impls/src/in_memory_store.rs
tankyleo: It sounds like we still need to fix this.

rust/server/src/util/config.rs
tankyleo, line 1: Let's remove all the changes in this file, and use a `cfg` flag to override the postgres backend configuration in `src/main.rs`. See how we use the `noop_authorizer`, we want to do something very similar for in-memory store as it will also be a developer tool and not something to be run in any production setting.

rust/server/src/main.rs
```diff
@@ -36,12 +37,24 @@ mod vss_service;
 
 fn main() {
 	let args: Vec<String> = std::env::args().collect();
+	if args.len() < 2 {
+		eprintln!("Usage: {} <config-file-path> [--in-memory]", args[0]);
+		std::process::exit(1);
+	}
```
tankyleo, line 43: A change of direction thank you: I prefer we use a `RUSTFLAGS="--cfg in-memory-store"` to set the backend, as it's a developer-only tool just like `noop_authorizer`. We can delete this as the postgres backend does not require a configuration file.

rust/server/src/main.rs
```diff
@@ -125,39 +138,50 @@ fn main() {
 			std::process::exit(-1);
 		});
 
-		let store: Arc<dyn KvStore> = if let Some(crt_pem) = config.tls_config {
```
tankyleo, line 128: Here let's use `cfg` flags to set the in-memory-store backend.

rust/server/vss-server-config.toml
```diff
@@ -3,6 +3,7 @@ bind_address = "127.0.0.1:8080" # Optional in TOML, can be overridden by env var
 # Maximum request body size in bytes. Can be set here or be overridden by env var 'VSS_MAX_REQUEST_BODY_SIZE'
 # Defaults to the maximum possible value of 1 GB if unset.
 # max_request_body_size = 1073741824
+store_type = "postgres" # "postgres" for using postgresql and "in-memory" for testing purposes
```
tankyleo, line 6: I prefer we make  no changes to the config file for a developer-only tool.

rust/README.md
```diff
@@ -26,6 +26,11 @@ cargo build --release
    ```
    cargo run server/vss-server-config.toml
    ```
+
+   **Note:** For testing purposes, you can pass `--in-memory` to use in-memory instead of PostgreSQL 
+   ```
+   cargo run -- server/vss-server-config.toml --in-memory
```
tankyleo, line 32: We'll have to update this to use `RUSTFLAGS` instead.

tankyleo (commented):
Thanks for your patience

PR #64, opened 2025-10-30
Rename `server` crate to `vss-server`

tankyleo:


rust/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [workspace]
 resolver = "2"
-members = ["server", "api", "impls", "auth-impls"]
-default-members = ["server"]
+members = ["versionedstorageserver", "api", "impls", "auth-impls"]
```
tnull: Should we maybe keep the `server` directory name, and just rename the crate?

PR #65, opened 2025-11-05
Return appropriate HTTP status codes in error responses

tnull:
In the deprecated Java version, we used to use appropriate HTTP status codes (400/409/etc): https://github.com/lightningdevkit/vss-server/blob/16874ec0638a2defc93d192508c8b962abc26926/java/app/src/main/java/org/vss/api/AbstractVssApi.java#L33-L52

However, in the Rust server we currently always return 500 / `INTERNAL_SERVER_ERROR. Here we fix this oversight.

### DIFF
--- a/rust/server/src/vss_service.rs
+++ b/rust/server/src/vss_service.rs
@@ -162,7 +162,7 @@ fn build_error_response(e: VssError) -> Response<Full<Bytes>> {
 			};
 			(status, error)
 		},
-		_ => {
+		VssError::InternalServerError(_) => {
 			let status = StatusCode::INTERNAL_SERVER_ERROR;
 			let error = ErrorResponse {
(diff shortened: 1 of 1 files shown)


rust/server/src/vss_service.rs
```diff
@@ -129,29 +129,50 @@ async fn handle_request<
...
+			let error = ErrorResponse {
+				error_code: ErrorCode::AuthException.into(),
+				message: msg.to_string(),
+			};
+			(status, error)
 		},
-		_ => ErrorResponse {
-			error_code: ErrorCode::InternalServerException.into(),
-			message: "Unknown Server Error occurred.".to_string(),
+		_ => {
```
tankyleo: LGTM if you don't mind let's be explicit about the error here I prefer to have them written out explicitly.

PR #66, opened 2025-11-05
Add `rustfmt` to CI to ensure formatting is checked

tnull:
.. we simply add a `cargo fmt` step to our Rust CI

### DIFF
--- a/.github/workflows/build-and-deploy-rust.yml
+++ b/.github/workflows/build-and-deploy-rust.yml
@@ -28,7 +28,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-
+      - name: Check formatting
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-deploy-rust.yml
```diff
@@ -28,7 +28,9 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
-
+      - name: Check formatting on Rust ${{ matrix.toolchain }}
+        if: matrix.check-fmt
```
tankyleo: When is this true ?

tnull: Ah, lol, good catch. It would be if I checked in the change. For simplicity I now simply dropped the line.

PR #72, opened 2025-12-06
Add option to verify JWT tokens in the HTTP Authorization header

tankyleo:


rust/server/src/util/config.rs
```diff
@@ -10,6 +10,7 @@ pub(crate) struct Config {
 pub(crate) struct ServerConfig {
 	pub(crate) host: String,
 	pub(crate) port: u16,
+	pub(crate) rsa_pub_file_path: Option<String>,
```
tnull: Shouldn't that rather be part of a new optional `JwtAuthConfig` section?

tnull (commented):
Generally looks good, some minor comments.

Do we see a chance of adding a test for the authorizer? Maybe with some fixtures?

rust/server/src/util/config.rs
tankyleo: See below @frnandu we now ask for the full PEM string, not the file that contains the string, either as an env var, or in the config.

tnull (approved):

PR #73, opened 2025-12-06
Add default database settting, and env var overrides

tankyleo:
Based on #72

### DIFF
--- a/rust/server/src/util/config.rs
+++ b/rust/server/src/util/config.rs
@@ -1,58 +1,159 @@
 use serde::Deserialize;
+use std::net::SocketAddr;
 
-#[derive(Deserialize)]
-pub(crate) struct Config {
-	pub(crate) server_config: ServerConfig,
-	pub(crate) jwt_auth_config: Option<JwtAuthConfig>,
-	pub(crate) postgresql_config: Option<PostgreSQLConfig>,
+const BIND_ADDR_VAR: &str = "VSS_BIND_ADDRESS";
+const JWT_RSA_PEM_VAR: &str = "VSS_JWT_RSA_PEM";
+const PSQL_USER_VAR: &str = "VSS_PSQL_USERNAME";
+const PSQL_PASS_VAR: &str = "VSS_PSQL_PASSWORD";
+const PSQL_ADDR_VAR: &str = "VSS_PSQL_ADDRESS";
+const PSQL_DB_VAR: &str = "VSS_PSQL_DEFAULT_DB";
+const PSQL_VSS_DB_VAR: &str = "VSS_PSQL_VSS_DB";
+const PSQL_TLS_VAR: &str = "VSS_PSQL_TLS";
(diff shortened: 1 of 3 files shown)


rust/server/src/util/config.rs
```diff
@@ -1,58 +1,142 @@
...
-			.expect("PostgreSQL database username must be provided in config or env var VSS_POSTGRESQL_USERNAME must be set.");
-		let password_env = std::env::var("VSS_POSTGRESQL_PASSWORD");
-		let password = password_env.as_ref()
-			.ok()
-			.or_else(|| self.password.as_ref())
-			.expect("PostgreSQL database password must be provided in config or env var VSS_POSTGRESQL_PASSWORD must be set.");
-
-		format!("postgresql://{}:{}@{}:{}", username, password, self.host, self.port)
-	}
+pub(crate) struct Configuration {
```
tnull: So now we have `Config` *and* `Configuration`? That seems rather confusing. Can they be merged, or at least one renamed appropriately to better distinguish their intended use?

rust/server/src/util/config.rs
```diff
@@ -1,58 +1,142 @@
...
-	let config: Config = toml::from_str(&config_str)?;
-	Ok(config)
+pub(crate) fn load_configuration(config_file_path: &str) -> Result<Configuration, String> {
+	let config_file = std::fs::read_to_string(config_file_path)
+		.map_err(|e| format!("Failed to read configuration file: {}", e))?;
+	let Config { server_config: ServerConfig { bind_address }, jwt_auth_config, postgresql_config } =
+		toml::from_str(&config_file)
+			.map_err(|e| format!("Failed to parse configuration file: {}", e))?;
+
+	macro_rules! read_env {
```
tnull: Do we really need macros for this? Maybe we can inline these or make them methods?

rust/server/src/util/config.rs
```diff
@@ -11,8 +11,10 @@ const PSQL_VSS_DB_VAR: &str = "VSS_PSQL_VSS_DB";
 const PSQL_TLS_VAR: &str = "VSS_PSQL_TLS";
 const PSQL_CERT_PEM_VAR: &str = "VSS_PSQL_CRT_PEM";
 
+// The structure of the toml config file. Any settings specified therein can be overriden by the corresponding
+// environment variable.
 #[derive(Deserialize, Default)]
-struct Config {
+struct TomlConfigTable {
```
tnull: nit: When you squash, can we drop the `Table` here?

tnull (commented):
> Thanks tnull addressed your comments let me know if I can rebase

Please go ahead and squash & rebase. While you're at it, you might address the one nit below.

tankyleo:
Rebased and squashed with the new `TomlConfig` change, and the change noted in the last commit message:
```
If a file is passed as an argument, and it does not exist, we now abort startup.
```

PR #76, opened 2025-12-14
Bound incoming request and add postgres service

enigbe:
### What this PR does
- Adds a PostgreSQL Docker service for the Rust server, removing the need for a local PostgreSQL installation
- Introduces a configurable maximum size limit for incoming request bodies, addressing a previously noted `TODO`
    - Operators can optionally set `maximum_request_body_size` in [`server_config`] or via environment variable
    - Defaults to 1 GB if not specified

### DIFF
--- a/rust/impls/src/postgres_store.rs
+++ b/rust/impls/src/postgres_store.rs
@@ -699,6 +699,10 @@ mod tests {
 	use super::{drop_database, DUMMY_MIGRATION, MIGRATIONS};
 	use crate::postgres_store::PostgresPlaintextBackend;
 	use api::define_kv_store_tests;
+	use api::kv_store::KvStore;
+	use api::types::{DeleteObjectRequest, GetObjectRequest, KeyValue, PutObjectRequest};
+
+	use bytes::Bytes;
 	use tokio::sync::OnceCell;
 	use tokio_postgres::NoTls;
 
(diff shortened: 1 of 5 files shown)


rust/server/src/vss_service.rs
```diff
@@ -18,6 +18,8 @@ use std::future::Future;
 use std::pin::Pin;
 use std::sync::Arc;
 
+const MAXIMUM_REQUEST_BODY_SIZE: u16 = 65_535;
```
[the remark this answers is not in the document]
tnull: As mentioned elsewhere: I guess a static upper bound is a good first step, but if we're really concerned about DoS we might need some dynamic rate limiting on a per-IP basis. Although then the question becomes how much of that should be considered the concern of the VSS service itself, and how much we'd just expect users to slap a load balancer/Cloudflare in front of the service to handle that for them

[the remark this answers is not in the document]
enigbe: I've made the configuration changes suggested here, capping at 20 MB for the maximum size.

rust/server/src/vss_service.rs
```diff
@@ -18,15 +18,45 @@ use std::future::Future;
 use std::pin::Pin;
 use std::sync::Arc;
 
+const MAXIMUM_REQUEST_BODY_SIZE: usize = 20 * 1024 * 1024;
+const DEFAULT_REQUEST_BODY_SIZE: usize = 10 * 1024 * 1024;
```
tnull: Not quite sure why we have two values now? Why have `DEFAULT` and a separate `MAXIMUM`?

rust/server/src/vss_service.rs
```diff
@@ -18,17 +18,45 @@ use std::future::Future;
 use std::pin::Pin;
 use std::sync::Arc;
 
-const MAXIMUM_REQUEST_BODY_SIZE: u16 = 65_535;
```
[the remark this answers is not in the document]
enigbe: I've rebased to avoid the repeated changes to the same lines.

rust/docker-compose.yml
```diff
@@ -0,0 +1,21 @@
+version: '3.8'
```
tankyleo: Thanks this like here emits this warning on my machine with a fresh docker let's delete it.
```
the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion
```

rust/api/src/kv_store_tests.rs
```diff
@@ -266,6 +267,29 @@ pub trait KvStoreTestSuite {
 		Ok(())
 	}
 
+	async fn put_should_succeed_with_maximum_supported_value_size() -> Result<(), VssError> {
```
tankyleo: Since this tests something specific to the `PostgresBackend` implementation, let's put this test inside `impls/src/postgres_store.rs`. This maximum body size of 1GiB is not part of the VSS protocol contract.

tankyleo (commented):
Thank your for the PR sorry for the delay on my part, a few bits, looking good !

rust/server/src/util/config.rs
```diff
@@ -94,11 +102,25 @@ pub(crate) fn load_configuration(config_file_path: Option<&str>) -> Result<Confi
...
+				format!("Unable to parse the maximum request body size environment variable: {}", e)
+			})
+		})
+		.transpose()?;
+	let maximum_request_body_size = read_config(
+		maximum_request_body_size_env,
+		max_request_body_size_config,
+		"VSS server maximum request body size",
+		MAX_REQUEST_BODY_SIZE,
+	)?;
```
tankyleo: Let's not make this parameter mandatory. In case it's not set, let's set it to `None` in the `Configuration` struct.

rust/server/src/vss_service.rs
```diff
@@ -18,15 +18,44 @@ use std::future::Future;
...
+	fn default() -> Self {
+		Self { maximum_request_body_size: MAXIMUM_REQUEST_BODY_SIZE }
+	}
+}
+
 #[derive(Clone)]
 pub struct VssService {
 	store: Arc<dyn KvStore>,
 	authorizer: Arc<dyn Authorizer>,
+	config: Arc<VssServiceConfig>,
```
tankyleo: I'd prefer we remove the `Arc<>` here, and do `#[derive(Clone, Copy)]` on `VssServiceConfig`

rust/server/src/main.rs
```diff
@@ -37,6 +37,16 @@ fn main() {
 			eprintln!("Failed to load configuration: {}", e);
 			std::process::exit(-1);
 		});
+	let vss_service_config = match &config.maximum_request_body_size {
+		Some(size) => match VssServiceConfig::new(*size) {
+			Ok(config) => Arc::new(config),
+			Err(e) => {
+				eprintln!("Configuration validation error: {}", e);
+				return;
```
tankyleo: One more thing, let's do an `std::process:exit(-1)` here like we do elsewhere

rust/server/vss-server-config.toml
```diff
@@ -1,5 +1,6 @@
 [server_config]
 bind_address = "127.0.0.1:8080" # Optional in TOML, can be overridden by env var `VSS_BIND_ADDRESS`
+maximum_request_body_size = 1073741824    # Optional in TOML:  maximum request body size in bytes capped at 1 GB, can be overriden by env var 'VSS_MAX_REQUEST_BODY_SIZE'
```
tankyleo: Let's keep this commented out to highlight to people that this parameter is optional; we'll default to 1GiB in case it is not set anywhere.

rust/impls/src/postgres_store.rs
```diff
@@ -695,10 +695,15 @@ where
 }
 
 #[cfg(test)]
+
```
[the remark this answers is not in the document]
enigbe: This has been removed.

rust/impls/src/postgres_store.rs
```diff
@@ -814,4 +818,70 @@ mod tests {
...
+	async fn supports_objects_up_to_non_large_object_threshold() {
+		let vss_db = "supports_objects_up_to_non_large_object_threshold";
+		let _ = drop_database(POSTGRES_ENDPOINT, DEFAULT_DB, vss_db, NoTls).await;
+
+		const MAXIMUM_SUPPORTED_VALUE_SIZE: usize = 1024 * 1024 * 1024;
+		const PROTOCOL_OVERHEAD_MARGIN: usize = 150;
+
+		// Construct entry that's for a field that's the maximum size of a non-"large_object" object
+		let large_value = vec![0u8; MAXIMUM_SUPPORTED_VALUE_SIZE - PROTOCOL_OVERHEAD_MARGIN];
+		let kv = KeyValue { key: "k1".into(), version: 0, value: Bytes::from(large_value) };
```
tankyleo: We'll need to clone `large_value` here

rust/server/src/main.rs
```diff
@@ -42,6 +42,16 @@ fn main() {
 			eprintln!("Failed to load configuration: {}", e);
 			std::process::exit(-1);
 		});
+	let vss_service_config = match &config.max_request_body_size {
+		Some(size) => match VssServiceConfig::new(*size) {
```
tankyleo: `usize` is `Copy`, so we can do `match config.max_request_body_size` and `VssServiceConfig::new(size)`

enigbe:
Thanks again for the review @tankyleo 
I've addressed the latest comments in 5cef3b6bdcebdff3430a0dc3669596c8e95722c2.

enigbe:
I've squashed the last fix-up.
Thank you.

PR #80, opened 2025-12-29
Add rust Dockerfile

### LINKED ISSUE #74: Dockerfile for rust version

frnandu:
fixes #74


rust/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+RUN cargo build --release --bin vss-server
+
+# Runtime stage
+FROM debian:bookworm-slim
+
+# Install runtime dependencies
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends \
+    ca-certificates \
+    libssl3 \
```
tankyleo: Wasn't aware of these dependencies are you sure these are needed ?

rust/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+    && rm -rf /var/lib/apt/lists/*
+
+WORKDIR /app
+
+# Copy the compiled binary from builder
+COPY --from=builder /build/target/release/vss-server /app/vss-server
+
+# Copy default configuration file
+#COPY server/vss-server-config.toml /app/vss-server-config.toml
```
tankyleo: For now this step is mandatory

tankyleo (commented):
Thanks ! can you make this one commit / remove the merge commit in the current PR ?

rust/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+
+# Copy workspace files
+COPY Cargo.toml Cargo.lock ./
+COPY rustfmt.toml ./
+
+# Copy all workspace members
+COPY server ./server
+COPY api ./api
+COPY impls ./impls
+COPY auth-impls ./auth-impls
```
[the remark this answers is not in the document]
frnandu: If we do `COPY . .` it would need to have a `.dockerignore` file with things like `target/**` and similar, otherwise it would unnecessarily copy compile stuff and make the build process take longer.
Let me know what you prefer.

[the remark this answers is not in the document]
tankyleo: Ah thanks sounds good let's leave it as is

tankyleo (commented):
As mentioned above, let's bring back `ca-certificates` for HTTPS requests.

Also squash all the commits into a single one.

frnandu:
Squashed, thanks!

tankyleo (commented):
Thanks tested on my machine looks good.

One last nitpick before merge: please set the title to `Add rust Dockerfile`, no need for the `chore` prefix thank you, I want to keep the commit log consistent.

frnandu:
> One last nitpick before merge: please set the title to `Add rust Dockerfile`, no need for the `chore` prefix thank you, I want to keep the commit log consistent.

done

frnandu:
ping?

tankyleo (dismissed):

tankyleo:
@frnandu I cleaned up the branch, used `fmar <fmar@fmar.dev>` for the author field on the dockerfile commit.

tnull thanks for taking a look I can't merge since I'm the last pusher.

rust/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+WORKDIR /app
+
+# Copy the compiled binary from builder
+COPY --from=builder /build/target/release/vss-server /app/vss-server
+
+# Copy default configuration file
+COPY server/vss-server-config.toml /app/vss-server-config.toml
+
+EXPOSE 8080
```
tnull, line 45: It seems by default this will only bind to localhost. Do we need to set `VSS_BIND_ADDRESS` to make this work outside of docker/reachable by other docker containers?

>   - [P2] Use container-reachable network defaults — /home/tnull/workspace/vss-server/rust/Dockerfile:40-40
>     When the image is run with its default CMD, e.g. docker run -p 8080:8080 ..., /app/vss-server-config.toml still binds VSS to 127.0.0.1 and points PostgreSQL at 127.0.0.1. Inside Docker those are container-loopback addresses, so the published port cannot reach the server and Postgres on
>     another container/host is unreachable unless users override both env vars. Consider shipping a Docker-specific config or setting Docker defaults such as VSS_BIND_ADDRESS=0.0.0.0:8080 and a network-reachable Postgres address.

tankyleo: Thanks made a couple more passes with codex, see below

tnull (approved):

PR #94, opened 2026-02-09
Drop jsonwebtoken dependency

tankyleo:
Fixes #92.

### DIFF
--- a/rust/auth-impls/Cargo.toml
+++ b/rust/auth-impls/Cargo.toml
@@ -5,18 +5,20 @@ edition = "2021"
 rust-version.workspace = true
 
 [features]
-jwt = [ "jsonwebtoken", "serde" ]
+jwt = [ "base64", "serde", "serde_json", "openssl" ]
 sigs = [ "bitcoin_hashes", "hex-conservative", "secp256k1" ]
 
 [dependencies]
-async-trait = "0.1.77"
 api = { path = "../api" }
-jsonwebtoken = { version = "9.3.0", optional = true, default-features = false, features = ["use_pem"] }
-serde = { version = "1.0.210", optional = true, default-features = false, features = ["derive"] }
-
+async-trait = "0.1.77"
+base64 = { version = "0.22.1", optional = true, default-features = false, features = ["std"] }
 bitcoin_hashes = { version = "0.19", optional = true, default-features = false }
 hex-conservative = { version = "1.0", optional = true, default-features = false }
+openssl = { version = "0.10.75", optional = true, default-features = false }
 secp256k1 = { version = "0.31", optional = true, default-features = false, features = [ "global-context" ] }
+serde = { version = "1.0.210", optional = true, default-features = false, features = ["derive"] }
+serde_json = { version = "1.0.149", optional = true, default-features = false, features = ["std"] }
 
 [dev-dependencies]
+jsonwebtoken = { version = "9.3.0", default-features = false, features = ["use_pem"] }
 tokio = { version = "1.38.0", default-features = false, features = ["rt-multi-thread", "macros"] }
--- a/rust/auth-impls/src/jwt.rs
+++ b/rust/auth-impls/src/jwt.rs
@@ -5,38 +5,139 @@
 use api::auth::{AuthResponse, Authorizer};
 use api::error::VssError;
 use async_trait::async_trait;
-use jsonwebtoken::{decode, Algorithm, DecodingKey, Validation};
+use base64::engine::general_purpose::{STANDARD, URL_SAFE_NO_PAD};
+use base64::Engine;
+use openssl::hash::MessageDigest;
+use openssl::pkey::Public;
+use openssl::pkey::{Id, PKey};
+use openssl::rsa::Rsa;
+use openssl::sign::Verifier;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
+use std::time::SystemTime;
+
+/// The tolerance used when validating the expiry claim of the token
+const CLOCK_SKEW_TOLERANCE_SEC: u64 = 60;
 
 /// A JWT based authorizer, only allows requests with verified 'JsonWebToken' signed by the given
 /// issuer key.
 ///
 /// Refer: https://datatracker.ietf.org/doc/html/rfc7519
 pub struct JWTAuthorizer {
-	jwt_issuer_key: DecodingKey,
+	jwt_issuer_key: PKey<Public>,
 }
 
 /// A set of Claims claimed by 'JsonWebToken'
 ///
 /// Refer: https://datatracker.ietf.org/doc/html/rfc7519#section-4
-#[derive(Serialize, Deserialize, Debug)]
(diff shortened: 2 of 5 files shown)


rust/auth-impls/Cargo.toml
```diff
@@ -5,18 +5,20 @@ edition = "2021"
...
-async-trait = "0.1.77"
 api = { path = "../api" }
-jsonwebtoken = { version = "9.3.0", optional = true, default-features = false, features = ["use_pem"] }
-serde = { version = "1.0.210", optional = true, default-features = false, features = ["derive"] }
-
+async-trait = "0.1.77"
+base64 = { version = "0.22.1", optional = true, default-features = false, features = ["std"] }
 bitcoin_hashes = { version = "0.19", optional = true, default-features = false }
 hex-conservative = { version = "1.0", optional = true, default-features = false }
+rsa = { version = "0.9.10", optional = true, default-features = false, features = ["sha2"] }
```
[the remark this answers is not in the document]
tankyleo: Thank you yes great idea

rust/auth-impls/src/jwt.rs
```diff
@@ -31,10 +36,22 @@ pub(crate) struct Claims {
 
 const BEARER_PREFIX: &str = "Bearer ";
 
+fn parse_public_key_pem(pem: &str) -> Result<PKey<Public>, String> {
```
tnull: Should we just return `()` or, if you prefer, a proper `enum` error type here? Using a string introduces some unnecessary allocations, and they can't  easily be handled. Same goes for `new` then ofc.

tnull (commented):
Thanks for the update, some minor comments, otherwise looks good.

However, I think `vss-server` in general gets into the realm where it could use some some fuzzing or proptest coverage to ensure that our implementations behave as expected, are safe, never panic, etc.

tankyleo:
Let me know what you think tnull, that last commit expands the scope of this PR a little where I add optional validation of `exp` claims, thought we could get this done along the way

rust/auth-impls/src/jwt.rs
```diff
@@ -5,41 +5,120 @@
...
+			.map_err(|_| VssError::AuthError(String::from("Claims json decode failed")))?;
+
+		// Check if the token is expired
+		if let Some(exp) = claims.exp {
+			let now = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH).unwrap().as_secs();
+			if now > exp.saturating_add(CLOCK_SKEW_TOLERANCE_SEC) {
+				return Err(VssError::AuthError(String::from("The token is expired")));
+			}
+		}
+
```
tnull, line 139: Codex:

> - High - rust/auth-impls/src/jwt.rs:103-118: aud is ignored. The old validator rejected tokens with an audience when no expected audience was configured; this now accepts tokens minted for another audience/service if signed by the same issuer.

rust/auth-impls/src/jwt.rs
```diff
@@ -5,41 +5,120 @@
...
+		// Decode the claims
+		let claims_json = URL_SAFE_NO_PAD
+			.decode(claims_base64)
+			.map_err(|_| VssError::AuthError(String::from("Claims base64 decode failed")))?;
+		let claims: Claims = serde_json::from_slice(&claims_json)
+			.map_err(|_| VssError::AuthError(String::from("Claims json decode failed")))?;
+
+		// Check if the token is expired
+		if let Some(exp) = claims.exp {
```
tnull: Codex:

> - High - rust/auth-impls/src/jwt.rs:39-40, rust/auth-impls/src/jwt.rs:111-116: JWTs without exp are now accepted. main used jsonwebtoken::Validation::new(Algorithm::RS256), which requires exp by default, so this creates non-expiring bearer tokens unless that policy change is
     intentional.

rust/auth-impls/src/jwt.rs
```diff
@@ -5,41 +5,120 @@
...
 	///
 	/// Refer: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.2
 	sub: String,
+	/// Expiration Time
+	exp: Option<u64>,
 }
 
 const BEARER_PREFIX: &str = "Bearer ";
 
+fn parse_public_key_pem(pem: &str) -> Result<PKey<Public>, ()> {
```
tnull, line 49: Codex:

> - Medium - rust/auth-impls/src/jwt.rs:45-54, rust/auth-impls/src/jwt.rs:94-98: RS256 is not tied to an RSA key. PKey::public_key_from_der accepts generic public keys, and OpenSSL verifies using the configured key type, so non-RSA keys are not rejected at setup.                         █

> - Medium - rust/auth-impls/src/jwt.rs:45-54: PEM parsing regressed to exact -----BEGIN PUBLIC KEY----- SPKI input. jsonwebtoken::from_rsa_pem also accepted common PKCS#1 -----BEGIN RSA PUBLIC KEY----- inputs, so some existing configs may fail to start.

tnull (commented):
Generally LGTM, but added some comments that Codex had for me. Seems they might be worth addressing even if not major?

> Let me know what you think tnull, that last commit expands the scope of this PR a little where I add optional validation of exp claims, thought we could get this done along the way

Sounds good, though, AFAIU Codex thinks the validation didn't use to be optional before?

tankyleo:
Thanks addressed review, also passed the code through codex :)

tankyleo:
> AFAIU Codex thinks the validation didn't use to be optional before?

Yes Codex is right, I misunderstood things given our old `Claims` struct only had a single `sub` member.

rust/auth-impls/src/jwt.rs
```diff
@@ -5,41 +5,142 @@
...
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
 pub(crate) struct Claims {
 	/// The "sub" (subject) claim identifies the principal that is the subject of the JWT.
 	/// The claims in a JWT are statements about the subject. This can be used as user identifier.
 	///
 	/// Refer: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.2
 	sub: String,
+	/// Expiration Time
+	exp: u64,
```
tnull, line 41: Codex:
```
  - rust/auth-impls/src/jwt.rs:40: exp is now deserialized as u64 in the application claims. The old production path decoded into Claims { sub } and let jsonwebtoken validate exp separately, which accepted JWT NumericDate values encoded as JSON floats. Tokens with valid fractional exp values
    will now fail with Claims json decode failed, so this is a compatibility regression.
```

Maybe that's fine?

tnull (commented):
Feel free to squash.

rust/auth-impls/src/jwt.rs
tankyleo: Yes I gave it some thought, would prefer to only add this if there is user demand.

tankyleo:
Squashed with no further changes

tnull (approved):

PR #96, opened 2026-04-02
Paginate by creation time instead of key order

benthecarman:
Clients often need to fetch recent payments or entries without iterating over the entire keyspace. Ordering by created_at lets callers retrieve the newest records first and stop early, which is not possible with lexicographic key ordering.

The page token now encodes (created_at, key) so the cursor remains unique even when multiple rows share the same timestamp. A composite index on (user_token, store_id, created_at, key) keeps the query efficient, and a migration back-fills any NULL created_at values and adds the NOT NULL constraint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,8 @@ integration with different cloud providers. As long as the API contract is imple
 effortlessly switch between different instances of VSS.
 
 VSS ships with a PostgreSQL implementation by default and can be hosted in your favorite infrastructure/cloud provider
-(AWS/GCP) and its backend storage can be switched with some other implementation for KeyValueStore if needed.
+(AWS/GCP). The backend storage can be switched with another implementation, but it must support ordering by creation
+time, a simple key-value store is not sufficient.
 
 ### Project Status
 
--- a/proto/vss.proto
+++ b/proto/vss.proto
@@ -210,13 +210,15 @@ message ListKeyVersionsRequest {
 // Server response for `ListKeyVersions` API.
 message ListKeyVersionsResponse {
 
-  // Fetched keys and versions.
+  // Fetched keys and versions, ordered by creation time (newest first).
   // Even though this API reuses the `KeyValue` struct, the `value` sub-field will not be set by the server.
   repeated KeyValue key_versions = 1;
 
   // `next_page_token` is a pagination token, used to retrieve the next page of results.
   // Use this value to query for next-page of paginated `ListKeyVersions` operation, by specifying
   // this value as the `page_token` in the next request.
+  //
+  // Following AIP-158 (https://google.aip.dev/158):
   //
   // If `next_page_token` is empty (""), then the "last page" of results has been processed and
   // there is no more data to be retrieved.
--- a/rust/api/src/kv_store_tests.rs
+++ b/rust/api/src/kv_store_tests.rs
@@ -53,6 +53,8 @@ macro_rules! define_kv_store_tests {
 		create_test!(list_should_honour_page_size_and_key_prefix_if_provided);
 		create_test!(list_should_return_zero_global_version_when_global_versioning_not_enabled);
 		create_test!(list_should_limit_max_page_size);
+		create_test!(list_should_return_results_ordered_by_creation_time);
+		create_test!(list_should_paginate_by_creation_time_with_prefix);
 	};
 }
 
@@ -393,12 +395,11 @@ pub trait KvStoreTestSuite {
 				},
 			};
 
-			if current_page.key_versions.is_empty() {
-				break;
-			}
-
 			all_key_versions.extend(current_page.key_versions);
-			next_page_token = current_page.next_page_token;
+			match current_page.next_page_token {
+				Some(token) if !token.is_empty() => next_page_token = Some(token),
+				_ => break,
+			}
 		}
 
 		if let Some(k1_response) = all_key_versions.iter().find(|kv| kv.key == "k1") {
@@ -444,13 +445,12 @@ pub trait KvStoreTestSuite {
 				},
 			};
 
-			if current_page.key_versions.is_empty() {
-				break;
-			}
-
 			assert!(current_page.key_versions.len() <= page_size as usize);
 			all_key_versions.extend(current_page.key_versions);
-			next_page_token = current_page.next_page_token;
+			match current_page.next_page_token {
+				Some(token) if !token.is_empty() => next_page_token = Some(token),
+				_ => break,
+			}
 		}
 
 		let unique_keys: std::collections::HashSet<String> =
@@ -490,18 +490,77 @@ pub trait KvStoreTestSuite {
 				Some(next_page_token) => ctx.list(Some(next_page_token), None, None).await?,
 			};
 
-			if current_page.key_versions.is_empty() {
-				break;
-			}
-
 			all_key_versions.extend(current_page.key_versions);
-			next_page_token = current_page.next_page_token;
+			match current_page.next_page_token {
+				Some(token) if !token.is_empty() => next_page_token = Some(token),
+				_ => break,
+			}
 		}
 
 		let unique_keys: std::collections::HashSet<String> =
 			all_key_versions.into_iter().map(|kv| kv.key).collect();
 		assert_eq!(unique_keys.len(), total_kv_objects as usize);
 		assert!(!unique_keys.contains(GLOBAL_VERSION_KEY));
+
+		Ok(())
+	}
+
+	async fn list_should_return_results_ordered_by_creation_time() -> Result<(), VssError> {
+		let kv_store = Self::create_store().await;
+		let ctx = TestContext::new(&kv_store);
+
+		ctx.put_objects(Some(0), vec![kv("z_first", "v1", 0)]).await?;
+		ctx.put_objects(Some(1), vec![kv("a_third", "v1", 0)]).await?;
+		ctx.put_objects(Some(2), vec![kv("m_second", "v1", 0)]).await?;
+
+		let page = ctx.list(None, None, None).await?;
+		assert_eq!(page.global_version, Some(3));
+		let keys: Vec<&str> = page.key_versions.iter().map(|kv| kv.key.as_str()).collect();
+
+		// Results should be in reverse creation order (newest first), not alphabetical.
+		assert_eq!(keys, vec!["m_second", "a_third", "z_first"]);
+
+		Ok(())
+	}
+
+	async fn list_should_paginate_by_creation_time_with_prefix() -> Result<(), VssError> {
+		let kv_store = Self::create_store().await;
+		let ctx = TestContext::new(&kv_store);
+
+		// Insert prefixed keys in reverse-alphabetical order with a page_size of 1
+		// to force multiple pages and verify cross-page ordering.
+		ctx.put_objects(Some(0), vec![kv("pfx_z", "v1", 0)]).await?;
+		ctx.put_objects(Some(1), vec![kv("pfx_a", "v1", 0)]).await?;
+		ctx.put_objects(Some(2), vec![kv("other", "v1", 0)]).await?;
+		ctx.put_objects(Some(3), vec![kv("pfx_m", "v1", 0)]).await?;
+
+		let mut next_page_token: Option<String> = None;
+		let mut all_keys: Vec<String> = Vec::new();
+
+		loop {
+			let current_page = match next_page_token.take() {
+				None => {
+					let page = ctx.list(None, Some(1), Some("pfx_".to_string())).await?;
+					assert_eq!(page.global_version, Some(4));
+					page
+				},
+				Some(token) => {
+					let page = ctx.list(Some(token), Some(1), Some("pfx_".to_string())).await?;
+					assert!(page.global_version.is_none());
+					page
+				},
+			};
+
+			assert!(current_page.key_versions.len() <= 1);
+			all_keys.extend(current_page.key_versions.into_iter().map(|kv| kv.key));
+			match current_page.next_page_token {
+				Some(token) if !token.is_empty() => next_page_token = Some(token),
+				_ => break,
+			}
+		}
+
+		// Should get prefixed keys in reverse creation order (newest first), excluding "other".
+		assert_eq!(all_keys, vec!["pfx_m", "pfx_a", "pfx_z"]);
 
 		Ok(())
 	}
@@ -524,16 +583,15 @@ pub trait KvStoreTestSuite {
 				None => ctx.list(None, None, None).await?,
 				Some(next_page_token) => ctx.list(Some(next_page_token), None, None).await?,
 			};
-			if current_page.key_versions.is_empty() {
-				break;
-			}
-
 			assert!(
 				current_page.key_versions.len() < vss_arbitrary_page_size_max as usize,
 				"Page size exceeds the maximum allowed size"
 			);
 			all_key_versions.extend(current_page.key_versions);
-			next_page_token = current_page.next_page_token;
+			match current_page.next_page_token {
+				Some(token) if !token.is_empty() => next_page_token = Some(token),
+				_ => break,
+			}
 		}
 
 		assert_eq!(all_key_versions.len(), total_kv_objects as usize);
--- a/rust/api/src/types.rs
+++ b/rust/api/src/types.rs
@@ -212,13 +212,15 @@ pub struct ListKeyVersionsRequest {
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct ListKeyVersionsResponse {
-	/// Fetched keys and versions.
+	/// Fetched keys and versions, ordered by creation time (newest first).
 	/// Even though this API reuses the `KeyValue` struct, the `value` sub-field will not be set by the server.
 	#[prost(message, repeated, tag = "1")]
 	pub key_versions: ::prost::alloc::vec::Vec<KeyValue>,
 	/// `next_page_token` is a pagination token, used to retrieve the next page of results.
 	/// Use this value to query for next-page of paginated `ListKeyVersions` operation, by specifying
 	/// this value as the `page_token` in the next request.
+	///
+	/// Following AIP-158 (<https://google.aip.dev/158>):
 	///
 	/// If `next_page_token` is empty (""), then the "last page" of results has been processed and
 	/// there is no more data to be retrieved.
--- a/rust/impls/src/migrations.rs
+++ b/rust/impls/src/migrations.rs
@@ -35,6 +35,8 @@ pub(crate) const MIGRATIONS: &[&str] = &[
 	    PRIMARY KEY (user_token, store_id, key)
 	);",
 	"ALTER TABLE vss_db DROP CONSTRAINT IF EXISTS vss_db_store_id_check;",
+	"ALTER TABLE vss_db ADD COLUMN IF NOT EXISTS sort_order BIGSERIAL NOT NULL UNIQUE;",
+	"CREATE INDEX IF NOT EXISTS idx_vss_db_sort_order ON vss_db (user_token, store_id, sort_order DESC) INCLUDE (key, version);",
 ];
 #[cfg(test)]
 pub(crate) const DUMMY_MIGRATION: &str = "SELECT 1 WHERE FALSE;";
--- a/rust/impls/src/postgres_store.rs
+++ b/rust/impls/src/postgres_store.rs
@@ -33,6 +33,22 @@ pub(crate) struct VssDbRecord {
 const KEY_COLUMN: &str = "key";
 const VALUE_COLUMN: &str = "value";
 const VERSION_COLUMN: &str = "version";
+const SORT_ORDER_COLUMN: &str = "sort_order";
+
+/// Page token is the `sort_order` value of the last item in the previous page,
+/// encoded as a decimal string.
+fn encode_page_token(sort_order: i64) -> String {
+	sort_order.to_string()
+}
+
+fn decode_page_token(token: &str) -> Result<i64, VssError> {
+	let invalid = || VssError::InvalidRequestError("Invalid page token".to_string());
+	let sort_order = token.parse::<i64>().map_err(|_| invalid())?;
+	if sort_order < 0 {
+		return Err(invalid());
+	}
+	Ok(sort_order)
+}
 
 /// The maximum number of key versions that can be returned in a single page.
 ///
@@ -659,25 +675,61 @@ where
 			global_version = Some(get_response.value.unwrap().version);
 		}
 
+		// When page_size is 0, we can decide how many items to return.
+		// Honor the 0 and just give the `global_version`.
+		if page_size == 0 {
+			return Ok(ListKeyVersionsResponse {
+				key_versions: vec![],
+				next_page_token: Some(String::new()),
+				global_version,
+			});
+		}
+
 		let limit = min(page_size, LIST_KEY_VERSIONS_MAX_PAGE_SIZE) as i64;
+		// Fetch one extra to determine if there are more pages.
+		let fetch_limit = limit + 1;
 
 		let conn = self.pool.get().await?;
 
-		let stmt = "SELECT key, version FROM vss_db WHERE user_token = $1 AND store_id = $2 AND key > $3 AND key LIKE $4 ORDER BY key LIMIT $5";
-
 		let key_like = format!("{}%", key_prefix.as_deref().unwrap_or_default());
-		let page_token_param = page_token.as_deref().unwrap_or_default();
-		let params: Vec<&(dyn tokio_postgres::types::ToSql + Sync)> =
-			vec![&user_token, &store_id, &page_token_param, &key_like, &limit];
 
-		let rows = conn
-			.query(stmt, &params)
-			.await
-			.map_err(|e| Error::new(ErrorKind::Other, format!("Query error: {}", e)))?;
+		let rows = if let Some(ref token) = page_token {
+			let page_sort_order = decode_page_token(token)?;
+			let stmt = "SELECT key, version, sort_order FROM vss_db WHERE user_token = $1 AND store_id = $2 AND sort_order < $3 AND key LIKE $4 AND key != $5 ORDER BY sort_order DESC LIMIT $6";
+			let params: Vec<&(dyn tokio_postgres::types::ToSql + Sync)> = vec![
+				&user_token,
+				&store_id,
+				&page_sort_order,
+				&key_like,
+				&GLOBAL_VERSION_KEY,
+				&fetch_limit,
+			];
+			conn.query(stmt, &params)
+				.await
+				.map_err(|e| Error::new(ErrorKind::Other, format!("Query error: {}", e)))?
+		} else {
+			let stmt = "SELECT key, version, sort_order FROM vss_db WHERE user_token = $1 AND store_id = $2 AND key LIKE $3 AND key != $4 ORDER BY sort_order DESC LIMIT $5";
+			let params: Vec<&(dyn tokio_postgres::types::ToSql + Sync)> =
+				vec![&user_token, &store_id, &key_like, &GLOBAL_VERSION_KEY, &fetch_limit];
+			conn.query(stmt, &params)
+				.await
+				.map_err(|e| Error::new(ErrorKind::Other, format!("Query error: {}", e)))?
+		};
+
+		let limit_usize = limit as usize;
+		let has_more = rows.len() > limit_usize;
+
+		let next_page_token = if has_more {
+			let last = &rows[limit_usize - 1];
+			let last_sort_order: i64 = last.get(SORT_ORDER_COLUMN);
+			Some(encode_page_token(last_sort_order))
+		} else {
+			Some(String::new())
+		};
 
 		let key_versions: Vec<_> = rows
 			.iter()
-			.filter(|&row| row.get::<&str, &str>(KEY_COLUMN) != GLOBAL_VERSION_KEY)
+			.take(limit_usize)
 			.map(|row| KeyValue {
 				key: row.get(KEY_COLUMN),
 				value: Bytes::new(),
@@ -685,22 +737,19 @@ where
 			})
 			.collect();
 
-		let mut next_page_token = Some("".to_string());
-		if !key_versions.is_empty() {
-			next_page_token = key_versions.get(key_versions.len() - 1).map(|kv| kv.key.to_string());
-		}
-
 		Ok(ListKeyVersionsResponse { key_versions, next_page_token, global_version })
 	}
 }
 
 #[cfg(test)]
 mod tests {
-	use super::{drop_database, DUMMY_MIGRATION, MIGRATIONS};
+	use super::{decode_page_token, drop_database, encode_page_token, DUMMY_MIGRATION, MIGRATIONS};
 	use crate::postgres_store::PostgresPlaintextBackend;
 	use api::define_kv_store_tests;
 	use api::kv_store::KvStore;
-	use api::types::{DeleteObjectRequest, GetObjectRequest, KeyValue, PutObjectRequest};
+	use api::types::{
+		DeleteObjectRequest, GetObjectRequest, KeyValue, ListKeyVersionsRequest, PutObjectRequest,
+	};
 
 	use bytes::Bytes;
 	use tokio::sync::OnceCell;
@@ -712,6 +761,28 @@ mod tests {
 	const MIGRATIONS_END: usize = MIGRATIONS.len();
 
 	static START: OnceCell<()> = OnceCell::const_new();
+
+	async fn put_test_key(
+		store: &PostgresPlaintextBackend, user_token: &str, store_id: &str, key: &str,
+		global_version: Option<i64>,
+	) {
+		store
+			.put(
+				user_token.to_string(),
+				PutObjectRequest {
+					store_id: store_id.to_string(),
+					global_version,
+					transaction_items: vec![KeyValue {
+						key: key.to_string(),
+						value: Bytes::from_static(b"v1"),
+						version: 0,
+					}],
+					delete_items: vec![],
+				},
+			)
+			.await
+			.unwrap();
+	}
 
 	define_kv_store_tests!(PostgresKvStoreTest, PostgresPlaintextBackend, {
 		let vss_db = "postgres_kv_store_tests";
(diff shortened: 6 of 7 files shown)


rust/impls/src/migrations.rs
```diff
@@ -35,6 +35,9 @@ pub(crate) const MIGRATIONS: &[&str] = &[
 	    PRIMARY KEY (user_token, store_id, key)
 	);",
 	"ALTER TABLE vss_db DROP CONSTRAINT IF EXISTS vss_db_store_id_check;",
+	"UPDATE vss_db SET created_at = COALESCE(last_updated_at, NOW()) WHERE created_at IS NULL;",
+	"ALTER TABLE vss_db ALTER COLUMN created_at SET NOT NULL;",
```
tankyleo: The `UPDATE` migration is there to make sure the second one does not fail ? Seems to me `created_at IS NULL` *should* never hit, as I am not aware of any VSS version that did not write timestamps, see 31562e01414cb05ba3388a52fe90c3b757ffe73f. Maybe there's some database out there that has missing values in that column.

rust/impls/src/migrations.rs
```diff
@@ -35,6 +35,9 @@ pub(crate) const MIGRATIONS: &[&str] = &[
...
+	"CREATE INDEX idx_vss_db_created_at ON vss_db (user_token, store_id, created_at, key);",
```
tankyleo: Here we prioritize read over write performance right ? Is that the correct priority for VSS ? I could see us prioritizing writes, which are on the hot path of user sending payments, and reads getting used rather rarely, ie in the device lost case.

rust/impls/src/postgres_store.rs
```diff
@@ -686,8 +712,10 @@ where
 			.collect();
 
 		let mut next_page_token = Some("".to_string());
-		if !key_versions.is_empty() {
-			next_page_token = key_versions.get(key_versions.len() - 1).map(|kv| kv.key.to_string());
+		if let Some(last_kv) = key_versions.last() {
+			let last_created_at =
+				rows[rows.len() - 1].get::<&str, chrono::DateTime<Utc>>(CREATED_AT_COLUMN);
```
tankyleo: nit: `rows.last().unwrap()`

rust/impls/src/postgres_store.rs
```diff
@@ -33,6 +33,25 @@ pub(crate) struct VssDbRecord {
 const KEY_COLUMN: &str = "key";
 const VALUE_COLUMN: &str = "value";
 const VERSION_COLUMN: &str = "version";
+const CREATED_AT_COLUMN: &str = "created_at";
+
+/// Page token is encoded as 20-char zero-padded epoch microseconds followed by the key.
+fn encode_page_token(created_at: &chrono::DateTime<Utc>, key: &str) -> String {
+	format!("{:020}{}", created_at.timestamp_micros(), key)
```
tankyleo: nit: on existing databases, the timestamps are in microsecond precision, so yea this makes sense: `2026-04-03 00:04:01.134768+00`. Can you add a quick comment here for this?

tankyleo (commented):
thanks !

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +508,59 @@ pub trait KvStoreTestSuite {
...
+		let mut all_keys: Vec<String> = Vec::new();
+
+		loop {
+			let current_page = match next_page_token.take() {
+				None => ctx.list(None, Some(1), Some("pfx_".to_string())).await?,
+				Some(token) => ctx.list(Some(token), Some(1), Some("pfx_".to_string())).await?,
+			};
+
+			if current_page.key_versions.is_empty() {
+				break;
```
tankyleo: nit: `assert_eq!(current_page.next_page_token, Some(String::new()));` ?

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +508,59 @@ pub trait KvStoreTestSuite {
...
+			}
+
+			assert!(current_page.key_versions.len() <= 1);
```
tankyleo, line 554: nit: `assert!(current_page.key_versions.len() == 1);`

rust/impls/src/postgres_store.rs
```diff
@@ -663,17 +682,24 @@ where
...
+		let rows = if let Some(ref token) = page_token {
+			let (page_created_at, page_key) = decode_page_token(token)?;
+			let stmt = "SELECT key, version, created_at FROM vss_db WHERE user_token = $1 AND store_id = $2 AND (created_at, key) > ($3, $4) AND key LIKE $5 ORDER BY created_at, key LIMIT $6";
+			let params: Vec<&(dyn tokio_postgres::types::ToSql + Sync)> =
+				vec![&user_token, &store_id, &page_created_at, &page_key, &key_like, &limit];
+			conn.query(stmt, &params)
+				.await
+				.map_err(|e| Error::new(ErrorKind::Other, format!("Query error: {}", e)))?
+		} else {
+			let stmt = "SELECT key, version, created_at FROM vss_db WHERE user_token = $1 AND store_id = $2 AND key LIKE $3 ORDER BY created_at, key LIMIT $4";
```
tankyleo: Question don't we want to return newest first ?

tankyleo (commented):
We'll want to update `types.rs` to require this ordering too

rust/impls/src/migrations.rs
benthecarman: Yeah, it was nullable before so need this to be safe.

rust/impls/src/migrations.rs
benthecarman: You're right that writes are the hot path and reads are rarer. In practice though, the extra index adds very little overhead per write. Happy to drop it if we'd rather optimize for writes.

As of now I don't think its recovery only though, currently ldk-node doesn't keep in local copy with vss store

benthecarman:
responded to review

rust/impls/src/migrations.rs
tankyleo: ah yes thanks

proto/vss.proto
```diff
@@ -210,7 +210,7 @@ message ListKeyVersionsRequest {
 // Server response for `ListKeyVersions` API.
 message ListKeyVersionsResponse {
 
-  // Fetched keys and versions.
+  // Fetched keys and versions, ordered by creation time (oldest first).
```
tankyleo: newest first

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +508,60 @@ pub trait KvStoreTestSuite {
...
+		// Insert keys in reverse-alphabetical order so creation order != key order.
+		ctx.put_objects(Some(0), vec![kv("z_first", "v1", 0)]).await?;
+		ctx.put_objects(Some(1), vec![kv("m_second", "v1", 0)]).await?;
+		ctx.put_objects(Some(2), vec![kv("a_third", "v1", 0)]).await?;
+
+		let page = ctx.list(None, None, None).await?;
+		let keys: Vec<&str> = page.key_versions.iter().map(|kv| kv.key.as_str()).collect();
+
+		// Results should be in reverse creation order (newest first), not alphabetical.
+		assert_eq!(keys, vec!["a_third", "m_second", "z_first"]);
```
tankyleo: let's rearrange the puts so the correct ordering does not match the alphabetical ordering

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +508,60 @@ pub trait KvStoreTestSuite {
...
+		let mut next_page_token: Option<String> = None;
+		let mut all_keys: Vec<String> = Vec::new();
+
+		loop {
+			let current_page = match next_page_token.take() {
+				None => ctx.list(None, Some(1), Some("pfx_".to_string())).await?,
+				Some(token) => ctx.list(Some(token), Some(1), Some("pfx_".to_string())).await?,
+			};
+
+			if current_page.key_versions.is_empty() {
```
tankyleo: I know this is pre-existing, and an issue with the other tests as well. But it seems totally within the API contract if a VSS returns a non-empty key-versions, with an empty page token. It would be weird for the client to then come back to the server, with an empty page token, when the VSS just told it to stop in the previous roundtrip.

I think this is a quick fix across all the tests to just check for the empty string to stop iteration, and not assert an empty key_versions list ?

proto/vss.proto
tankyleo: To confirm here, the server is free to choose any ordering they would like in case two keys have the same timestamp.

rust/impls/src/postgres_store.rs
```diff
@@ -884,4 +914,27 @@ mod tests {
...
+	#[test]
+	fn page_token_rejects_short_input() {
+		assert!(decode_page_token("tooshort").is_err());
+		assert!(decode_page_token("").is_err());
+	}
+
+	#[test]
+	fn page_token_rejects_non_numeric_prefix() {
+		assert!(decode_page_token("abcdefghijklmnopqrstkey").is_err());
+	}
```
tankyleo, line 1169: Can we ask claude to write a test that asserts the descendant key ordering in case two records share the same timestamp ?

tankyleo:
please request review when you are ready so i don't miss this PR

proto/vss.proto
benthecarman: We have the server sort by key after timestamp. I'll add that here

tankyleo: Yes, but I don't this is a requirement on the VSS API right ? Ie clients don't care what ordering is picked beyond newest first.

rust/impls/src/migrations.rs
tankyleo: Yes good to keep the index, I think we can easily drop it in a future migration. Feel free to resolve this.

proto/vss.proto
benthecarman: I think it's better to me more explicit than less. We have to define a tiebreaker anyways so may as well put it in the docs

proto/vss.proto
```diff
@@ -210,20 +210,16 @@ message ListKeyVersionsRequest {
...
-  // Fetched keys and versions.
+  // Fetched keys and versions, ordered by creation time (newest first) ties broken by key (alphabetically).
   // Even though this API reuses the `KeyValue` struct, the `value` sub-field will not be set by the server.
   repeated KeyValue key_versions = 1;
 
   // `next_page_token` is a pagination token, used to retrieve the next page of results.
   // Use this value to query for next-page of paginated `ListKeyVersions` operation, by specifying
   // this value as the `page_token` in the next request.
   //
-  // If `next_page_token` is empty (""), then the "last page" of results has been processed and
```
tnull, line 221: No, I think we were intentionally following protobuf's/Google's best practices here. https://google.aip.dev/158 states:

> - Response messages for collections should define a string next_page_token field, providing the user with a page token that may be used to retrieve the next page.
>     - The field containing pagination results should be the first field in the message and have a field number of 1. It should be a repeated field containing a list of resources constituting a single page of results.
>     - If the end of the collection has been reached, the next_page_token field must be empty. This is the only way to communicate "end-of-collection" to users.
>     - If the end of the collection has not been reached (or if the API can not determine in time), the API must provide a next_page_token.

IMO would be good to revert and include this context in the docs.

benthecarman: Okay updated

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +507,104 @@ pub trait KvStoreTestSuite {
 		Ok(())
 	}
 
+	async fn list_should_return_results_ordered_by_creation_time() -> Result<(), VssError> {
+		let kv_store = Self::create_store().await;
+		let ctx = TestContext::new(&kv_store);
+
+		ctx.put_objects(None, vec![kv("z_first", "v1", 0)]).await?;
+		ctx.put_objects(None, vec![kv("a_third", "v1", 0)]).await?;
+		ctx.put_objects(None, vec![kv("m_second", "v1", 0)]).await?;
```
tankyleo: Let's bring back the global versions ?

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +507,104 @@ pub trait KvStoreTestSuite {
...
+	async fn list_should_paginate_by_creation_time_with_prefix() -> Result<(), VssError> {
+		let kv_store = Self::create_store().await;
+		let ctx = TestContext::new(&kv_store);
+
+		// Insert prefixed keys in reverse-alphabetical order with a page_size of 1
+		// to force multiple pages and verify cross-page ordering.
+		ctx.put_objects(None, vec![kv("pfx_z", "v1", 0)]).await?;
+		ctx.put_objects(None, vec![kv("pfx_a", "v1", 0)]).await?;
+		ctx.put_objects(None, vec![kv("other", "v1", 0)]).await?;
+		ctx.put_objects(None, vec![kv("pfx_m", "v1", 0)]).await?;
```
tankyleo: Same here I believe we can bring back the global versions

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +507,104 @@ pub trait KvStoreTestSuite {
...
+
+		loop {
+			let current_page = match next_page_token.take() {
+				None => ctx.list(None, Some(1), Some("pfx_".to_string())).await?,
+				Some(token) => ctx.list(Some(token), Some(1), Some("pfx_".to_string())).await?,
+			};
+
+			all_keys.extend(current_page.key_versions.into_iter().map(|kv| kv.key));
+			match current_page.next_page_token {
+				Some(token) if !token.is_empty() => next_page_token = Some(token),
```
tankyleo, line 557: Here we still want to assert that the length of all the returned pages was 1 or 0 (see "The server may further constrain the maximum number of results returned in a single page." from `types.rs`).

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +507,104 @@ pub trait KvStoreTestSuite {
...
+			}
+		}
+
+		// Should get prefixed keys in reverse creation order (newest first), excluding "other".
+		assert_eq!(all_keys, vec!["pfx_m", "pfx_a", "pfx_z"]);
+
+		Ok(())
+	}
+
+	async fn list_should_return_empty_page_token_when_exact_fit() -> Result<(), VssError> {
```
tankyleo: Let's move this test to the `postgres` implementation, as this is not a requirement of all implementations of the server-side of the VSS API.

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +507,104 @@ pub trait KvStoreTestSuite {
...
+		let kv_store = Self::create_store().await;
+		let ctx = TestContext::new(&kv_store);
+
+		let page_size = 5;
+		for i in 0..page_size {
+			ctx.put_objects(None, vec![kv(&format!("k{}", i), "v1", 0)]).await?;
+		}
+
+		let page = ctx.list(None, Some(page_size), None).await?;
+		assert_eq!(page.key_versions.len(), page_size as usize);
```
tankyleo: Given the VSS API definition, you cannot make this assumption here.

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +507,104 @@ pub trait KvStoreTestSuite {
...
+		assert_eq!(
+			page.next_page_token.as_deref(),
+			Some(""),
+			"expected empty token when results fit exactly in one page"
```
tankyleo: Same here I think this is an assumption we can make for our specific postgres implementation, but this is not a requirement of all VSS servers.

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +507,104 @@ pub trait KvStoreTestSuite {
...
+			first_page.next_page_token.as_ref().map_or(false, |t| !t.is_empty()),
+			"expected a non-empty page token when more items exist"
+		);
+
+		let second_page = ctx.list(first_page.next_page_token, Some(page_size), None).await?;
+		assert_eq!(second_page.key_versions.len(), 1);
+		assert_eq!(
+			second_page.next_page_token.as_deref(),
+			Some(""),
+			"expected empty token on the last page, got {:?}",
```
tankyleo: This is true for our VSS implementation, but not a requirement of all VSS-server implementations.

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +507,104 @@ pub trait KvStoreTestSuite {
...
+		let kv_store = Self::create_store().await;
+		let ctx = TestContext::new(&kv_store);
+
+		let page_size = 5;
+		for i in 0..page_size + 1 {
+			ctx.put_objects(None, vec![kv(&format!("k{}", i), "v1", 0)]).await?;
+		}
+
+		let first_page = ctx.list(None, Some(page_size), None).await?;
+		assert_eq!(first_page.key_versions.len(), page_size as usize);
```
tankyleo: Given the VSS API, the page could have length 0 here. As long as the page token is not empty, the client would be expected to make another request with the new page token.

rust/impls/src/postgres_store.rs
```diff
@@ -660,47 +680,73 @@ where
 		}
 
 		let limit = min(page_size, LIST_KEY_VERSIONS_MAX_PAGE_SIZE) as i64;
+		// Fetch one extra to determine if there are more pages.
+		let fetch_limit = limit + 1;
```
tankyleo, line 690: With this you are looking to eliminate the last roundtrip, ie set the page token to the empty string on the last non-empty page, instead of returning some non-empty pagen token with the last page, and the client making another roundrip before the server returns the empty string ?

I did like the simplicity of the prior version, even though it required an extra roundtrip. Basically just set the cursor to the last key of the page. This is within the VSS API I'm quite sure.

WDYT ?

tankyleo (commented):
Rushed some comments out, will be back after lunch.

proto/vss.proto
tankyleo: Let me see this defines an API constraint right, in addition to just documentation ? For sure I see the use for documentation here, but I don't want clients to start relying on this behavior.

tankyleo:
Done with this pass here, just needed to add the comment about the VSS server API constraint

G8XSU:
Adding some historical context, VSS was purposefully built to be storage engine agnostic. The VSS protocol/API/data-model doesn't enforce a specific underlying storage database.

We can totally revisit this requirement and say VSS only works with postgres-like relational databases. But extending pagination to order by creation_time limits the capability to use any modern pure KV store as an underlying database engine like DynamoDB or CosmosDB.

I understand the need to efficiently paginate through large result sets. A couple of alternatives that preserve storage engine agnosticism:
  - Clients could use ordered key IDs like UUIDv7, which gives you time-based ordering for free
  via lexicographic sort on the key itself.
  - Clients could scope their list_key_versions calls to known key prefixes to keep result sets
  manageable.
  - Cursor-based pagination on the key (lexicographic ordering) works on every KV store without
  requiring timestamps.

rust/api/src/kv_store_tests.rs
benthecarman: I'm confused what you're saying here. This is the first page we are requesting, the only way we'd a length of 0 would be if we had no items.

rust/impls/src/postgres_store.rs
benthecarman: I like this better, saves a round trip for little over head. Here the round trips can actually be kinda expensive because its a network call unlike for our local kv store backends, however even for those we are doing this logic as well.

benthecarman:
Fixed @tankyleo comments about tests

tankyleo:
@G8XSU Thank you for the feedback, I'll be considering the tradeoffs over the next few days.

Was wondering do you have an email where I can reach you ? Feel free to ping me at "hello at leonash dot net" I was wondering if we could get the vss-client crate at https://crates.io/crates/vss-client from you.

benthecarman:
@G8XSU my perspective

> Clients could use ordered key IDs like UUIDv7, which gives you time-based ordering for free
via lexicographic sort on the key itself.

in theory, yes, but ldk-node isn't using this and migrating would be a huge lift

> Clients could scope their list_key_versions calls to known key prefixes to keep result sets
manageable.

ldk-node uses txid/payment_hash as key so we can't really use prefixes

> Cursor-based pagination on the key (lexicographic ordering) works on every KV store without
requiring timestamps.

Yes, but then the pagination isn't really useful anymore. You just end up getting a random set of payments rather an actual ordered list.

> We can totally revisit this requirement and say VSS only works with postgres-like relational databases. But extending pagination to order by creation_time limits the capability to use any modern pure KV store as an underlying database engine like DynamoDB or CosmosDB.

imo postgres scales really well and you're scaling past something like postgres then it's likely you aren't using the off the shelf solution anymore and will have a whole team to manage things like stuff. And being able to sort by something useful rather than just lexicographical is worth it

rust/impls/src/postgres_store.rs
tankyleo: thank you ok

rust/api/src/kv_store_tests.rs
tankyleo: Just that I think a VSS-server is within bounds if it returns a response with an empty list, but a non-empty token. The client would be expected to continue asking for pages.

I'm mostly going from this line in the docs of `ListKeyVersionRequest`:
```
        /// `page_size` is used by clients to specify the maximum number of results that can be returned by
        /// the server.
        /// The server may further constrain the maximum number of results returned in a single page.
        /// If the `page_size` is 0 or not set, the server will decide the number of results to be returned.
        #[prost(int32, optional, tag = "3")]
        pub page_size: ::core::option::Option<i32>,
```

TLDR can't assume the page you get back is the same length as the page size in your `ListKeyVersionsRequest`

rust/api/src/types.rs
```diff
@@ -212,14 +212,16 @@ pub struct ListKeyVersionsRequest {
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct ListKeyVersionsResponse {
-	/// Fetched keys and versions.
+	/// Fetched keys and versions, ordered by creation time (newest first) ties broken by key (alphabetically).
```
tankyleo: We discussed this offline @benthecarman do you think we can drop this "ties broken alphabetically" at this point ?

benthecarman: fixed

G8XSU:
> in theory, yes, but ldk-node isn't using this and migrating would be a huge lift

Yes I understand, but vss wasn't built with just keeping ldk-node payment-store in mind. Maybe it is not the right for payment-store currently.

But tying vss interface to be relational db specific is a 1-way door decision, and such interface change hasn't been made to either ldk-node or ldk iiuc.

> Yes, but then the pagination isn't really useful anymore. You just end up getting a random set of payments rather an actual ordered list.

Hmm.. maybe, but why are we only thinking about payment store? Payment store requires PaginatedKvStore interface and on top of that it needs creation time based sorting. Imo the right solution would still be changing payment-storage to use something like uuidv7 so that it can work with any key based paginated store efficiently.
Instead of compensating for that in vss. 

> imo postgres scales really well and you're scaling past something like postgres then it's likely you aren't using the off the shelf solution anymore and will have a whole team to manage things like stuff. And being able to sort by something useful rather than just lexicographical is worth it

It is not about scaling at all but operational flexibility, can we get the change to sort by creation_time through ldk-storage interface? If not, then the similar reasons apply here as well i.e. clients want to use the storage that they want to use. 

> it's likely you aren't using the off the shelf solution anymore

It's not about this server code in repo but interoperable protocol and api model.
In the process, we have lost any chance of inter-operable vss protocol with hosted 3p providers, otherwise everyone must use relational db, they can't cost arbitrage or manage their operations in their own way.

I understand it is limiting to some extent but I just wanted to flag the reason for having a conservative api/protocol approach here.

rust/api/src/kv_store_tests.rs
G8XSU: > TLDR can't assume the page you get back is the same length as the page size in your ListKeyVersionsRequest

This is true. 

> if it returns a response with an empty list, but a non-empty token.

But there is no reason to respond with empty list and non-empty pagination-token. (I don't think this should happen, can add a kvstore test/assert for it if doesn't exist already.)

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +505,66 @@ pub trait KvStoreTestSuite {
...
+		ctx.put_objects(Some(0), vec![kv("z_first", "v1", 0)]).await?;
+		ctx.put_objects(Some(1), vec![kv("a_third", "v1", 0)]).await?;
+		ctx.put_objects(Some(2), vec![kv("m_second", "v1", 0)]).await?;
+
+		let page = ctx.list(None, None, None).await?;
+		assert_eq!(page.global_version, Some(3));
+		let keys: Vec<&str> = page.key_versions.iter().map(|kv| kv.key.as_str()).collect();
+
+		// Results should be in reverse creation order (newest first), not alphabetical.
+		assert_eq!(keys, vec!["m_second", "a_third", "z_first"]);
```
tankyleo, line 521: Since this is for VSS API tests, you can't assume you'll get all the keys back in a single go. It's the server's decision how many keys you get per page.

rust/api/src/types.rs
tankyleo: Let's remove this "ties broken by key (alphabetically)" too

rust/impls/src/migrations.rs
tankyleo: Looks like we could append an `INCLUDE (version)` here WDYT ? `list_key_versions` could run just by looking at the index, without looking at the main table.

rust/impls/src/postgres_store.rs
```diff
@@ -660,47 +680,73 @@ where
...
+			conn.query(stmt, &params)
+				.await
+				.map_err(|e| Error::new(ErrorKind::Other, format!("Query error: {}", e)))?
+		};
+
+		let limit_usize = limit as usize;
+		let has_more = rows.len() > limit_usize;
+
+		let next_page_token = if has_more {
+			let last = &rows[limit_usize - 1];
```
tankyleo, line 723: This can overflow + panic if a client maliciously sets the limit to 0.

rust/impls/src/postgres_store.rs
```diff
@@ -884,4 +952,162 @@ mod tests {
...
+						store_id: "store".to_string(),
+						page_token: None,
+						page_size: Some(5),
+						key_prefix: None,
+					},
+				)
+				.await
+				.unwrap();
+
+			assert_eq!(resp.key_versions.len(), 5);
```
tankyleo, line 1086: let's also assert on the value of global version returned here too

rust/impls/src/postgres_store.rs
```diff
@@ -884,4 +952,162 @@ mod tests {
...
+						store_id: "store".to_string(),
+						page_token: None,
+						page_size: Some(5),
+						key_prefix: None,
+					},
+				)
+				.await
+				.unwrap();
+
+			assert_eq!(first_page.key_versions.len(), 5);
```
tankyleo, line 1124: same here let's assert on the value of global_version

tankyleo (commented):
In case we decide to proceed with this approach

rust/api/src/kv_store_tests.rs
benthecarman: i mean, can we at least assume the page size is >=3 for this test lol?

rust/impls/src/postgres_store.rs
benthecarman: should we just return an error for 0 page size, seems pointless

tankyleo: can we return early with a hardcoded `ListKeyVersionResponse`  right after `let limit` ? just grab the global_version like it is currently on master ?

rust/api/src/kv_store_tests.rs
tankyleo: sounds good haha it's just a test

rust/impls/src/migrations.rs
benthecarman: I didn't know you could do that, cool

tankyleo: Google AI suggested this :)

benthecarman: TIL someone uses google ai :joy:

rust/api/src/kv_store_tests.rs
```diff
@@ -393,12 +395,11 @@ pub trait KvStoreTestSuite {
 				},
 			};
 
-			if current_page.key_versions.is_empty() {
```
tnull, line 396: Why was this removed? (same below)

rust/api/src/kv_store_tests.rs
```diff
@@ -393,12 +395,11 @@ pub trait KvStoreTestSuite {
...
-				break;
-			}
-
 			all_key_versions.extend(current_page.key_versions);
-			next_page_token = current_page.next_page_token;
+			match current_page.next_page_token {
+				Some(token) if !token.is_empty() => next_page_token = Some(token),
```
tnull, line 400: Also a bit dubious why we need these test changes?

rust/impls/src/postgres_store.rs
```diff
@@ -659,48 +679,83 @@ where
...
+			let last_created_at = last.get::<&str, chrono::DateTime<Utc>>(CREATED_AT_COLUMN);
+			let last_key: String = last.get(KEY_COLUMN);
+			Some(encode_page_token(&last_created_at, &last_key))
+		} else {
+			Some(String::new())
+		};
 
 		let key_versions: Vec<_> = rows
 			.iter()
-			.filter(|&row| row.get::<&str, &str>(KEY_COLUMN) != GLOBAL_VERSION_KEY)
```
tnull, line 680: We don't we need that filter anymore?

rust/impls/src/postgres_store.rs
```diff
@@ -33,6 +33,26 @@ pub(crate) struct VssDbRecord {
 const KEY_COLUMN: &str = "key";
 const VALUE_COLUMN: &str = "value";
 const VERSION_COLUMN: &str = "version";
+const CREATED_AT_COLUMN: &str = "created_at";
+
+/// Page token is encoded as 20-char zero-padded epoch microseconds followed by the key.
+fn encode_page_token(created_at: &chrono::DateTime<Utc>, key: &str) -> String {
```
tnull: Is this really worth it? Are our API guarantees so strict that we need to guarantee strict monotonicity vs. 'just' monotonicity? Maybe it would be simpler to just fallback to lexicographic ordering as a tie breaker?  

If we really do want to add semantics to the page cursor, it would be better to also include a versioning byte to make sure we can change the semantics going forward.

rust/api/src/types.rs
tnull, line 199: This directly previously-existing API contract now directly conflicts with the assumption below:
```rust
// A page_size of 0 means the client only wants the global_version.
```

rust/impls/src/migrations.rs
```diff
@@ -35,6 +35,9 @@ pub(crate) const MIGRATIONS: &[&str] = &[
 	    PRIMARY KEY (user_token, store_id, key)
 	);",
 	"ALTER TABLE vss_db DROP CONSTRAINT IF EXISTS vss_db_store_id_check;",
+	"UPDATE vss_db SET created_at = COALESCE(last_updated_at, NOW()) WHERE created_at IS NULL;",
+	"ALTER TABLE vss_db ALTER COLUMN created_at SET NOT NULL;",
+	"CREATE INDEX idx_vss_db_created_at ON vss_db (user_token, store_id, created_at, key) INCLUDE (version);",
```
tnull: Codex:

> The new Postgres index does not actually match the query’s sort order.
>      The query orders by created_at DESC, key ASC in rust/impls/src/postgres_store.rs:701 and :715, but the migration adds CREATE INDEX ... (user_token, store_id, created_at, key) in rust/impls/src/migrations.rs:40, which is effectively ASC, ASC. A plain btree can satisfy ASC, ASC or the
>      full reverse DESC, DESC; it does not naturally satisfy the mixed direction DESC, ASC. Postgres may still need a sort or incremental sort, so the commit message’s claim that this index "keeps the query efficient" is not reliably true.

tnull (commented):
While to a degree I do share the concerns about further restricting the set of possible (future) backends for VSS, I have to begrudgingly say that I don't see a better alternative, e.g.:

> I understand the need to efficiently paginate through large result sets. A couple of alternatives that preserve storage engine agnosticism:
> 
> - Clients could use ordered key IDs like UUIDv7, which gives you time-based ordering for free
> via lexicographic sort on the key itself.

Requiring this from the client a) would be very restrictive, but even more importantly conflicts with many, many design choices we made so far. In an ideal world we could discuss at length what would/could make sense, but realistically that won't happen.

> - Clients could scope their list_key_versions calls to known key prefixes to keep result sets
> manageable.

We don't have a set of known prefixes as we intentionally obfuscate namespaces.

> - Cursor-based pagination on the key (lexicographic ordering) works on every KV store without
> requiring timestamps.

This would of course work, but then in practice it would mean we'd need to paginate through all results, create a huge allocated `Vec`, sort it, and then return a subset. At that point it's unclear why we add pagination in the first place.


That said, it would still be great if we found a better solution that meets our requirements *and* doesn't require us to further restrict our backend choices going forward.

rust/api/src/kv_store_tests.rs
tankyleo: According to the API docs, the real test for whether we've finished pagination is not "is key_versions empty?", but rather, "is next_page_token empty?"

See:
```
        /// If `next_page_token` is empty (""), then the "last page" of results has been processed and
        /// there is no more data to be retrieved.
        ///
        /// If `next_page_token` is not empty, it does not necessarily mean that there is more data in the
        /// result set. The only way to know when you have reached the end of the result set is when
        /// `next_page_token` is empty.
```

rust/api/src/kv_store_tests.rs
tankyleo: See the response above. Here we also make the assumption that if the server returns `None` for the page token, this means we've reached the end of pagination. This is not explicitly documented in the API docs.

rust/impls/src/postgres_store.rs
tankyleo: The sql query now has a `key != $4` clause that filters out the global version.

rust/api/src/types.rs
tankyleo: I don't quite see the conflict here, the server can decide how many results to return. Yes the comment may be off as it's not quite "the client only wants the global version", but rather "we get to decide."

I asked originally to implement the same behavior we currently have on master, but happy to revisit this decision.

rust/impls/src/migrations.rs
benthecarman: codex my goat

rust/impls/src/postgres_store.rs
benthecarman: > Maybe it would be simpler to just fallback to lexicographic ordering as a tie breaker?

we currently do.

> If we really do want to add semantics to the page cursor, it would be better to also include a versioning byte to make sure we can change the semantics going forward.

I added a version byte so we have safety going in to the future.

rust/api/src/types.rs
tnull: Right, but if it's not part of the API guarantees, isn't it odd to lean on/assume that the client only wants global version. At the very least the comment is misleading, IMO.

rust/impls/src/postgres_store.rs
```diff
@@ -33,6 +33,33 @@ pub(crate) struct VssDbRecord {
...
+	// Uses microsecond precision to match the timestamp precision stored in Postgres.
+	format!("{}:{:020}:{}", CURRENT_PAGE_TOKEN_VERSION, created_at.timestamp_micros(), key)
+}
+
+fn decode_page_token(token: &str) -> Result<(chrono::DateTime<Utc>, String), VssError> {
+	let invalid = || VssError::InvalidRequestError("Invalid page token".to_string());
+	let rest = match token.chars().next() {
+		Some(CURRENT_PAGE_TOKEN_VERSION) => {
+			token.get(2..).filter(|_| token.as_bytes().get(1) == Some(&b':')).ok_or_else(invalid)?
```
tnull: Given we already introduced a separator (`:`), shouldn't we rather `split` instead of assuming the second character is where the timestamp starts?

benthecarman: done

tankyleo (commented):
LGTM

> Right, but if it's not part of the API guarantees, isn't it odd to lean on/assume that the client only wants global version. At the very least the comment is misleading, IMO.

Let's just address this comment from tnull thank you, perhaps something like "page 0 means we get to decide"

benthecarman:
Updated the comment about 0 page size

rust/api/src/kv_store_tests.rs
tnull: Okay. Nit: could have happened in a separate commit, describing the rationale in the message.

proto/vss.proto
```diff
@@ -210,14 +210,16 @@ message ListKeyVersionsRequest {
 // Server response for `ListKeyVersions` API.
 message ListKeyVersionsResponse {
 
-  // Fetched keys and versions.
+  // Fetched keys and versions, ordered by creation time (newest first).
```
tnull, line 213: Remind me, is there any particular reason we need to use time for this? Wouldn't we get around the need for a tie breaker if we'd simply use a monotonically increasing atomic counter instead  (e.g., a Postgres `BIGSERIAL` column) ? Then we'd be certain than each entry has a unique value?

tnull (commented):
Generally fine by me I think, one comment.

Aside from that, I do however wonder if we should finally add a protocol version to VSS? While this is not an API breaking change, clients will have no good idea whether they can expect pagination to work when they connect to a particular backend? 

@tankyleo Any thoughts?

proto/vss.proto
tnull: Claude:
```
 Monotonic counter vs. creation_time

  Your intuition is right — a monotonically increasing counter (e.g., a Postgres BIGSERIAL column) would be strictly simpler here. The current
  approach has to deal with:

  - Tie-breaking: The compound condition (created_at < $3 OR (created_at = $3 AND key > $4)) in the SQL query
  - A composite page token: Encoding both the timestamp and the key (0:<micros>:<key>)
  - A composite index: (user_token, store_id, created_at DESC, key ASC) INCLUDE (version)
  - A dedicated test just for the tie-breaking behavior

  With an auto-incrementing counter (say row_id BIGSERIAL), all of that collapses to:
  - WHERE row_id < $cursor ORDER BY row_id DESC LIMIT $N
  - Page token = just the counter value
  - Simple single-column addition to the index
  - No tie-breaking needed since values are unique by definition

  The only argument for creation_time would be if it carried semantic meaning the client cares about (e.g., "show me keys created after X"). But
  looking at the ListKeyVersionsResponse, created_at isn't exposed to the client — it's purely an internal ordering mechanism. So you're paying the
   complexity tax of timestamps without getting the semantic benefit.
```

tankyleo: Thank you tnull for raising this point. I did some back-and-forth with claude, and yea this is interesting !

My one question at this point is backwards compat. Do you have thoughts on this point ? I'm thinking if keys created before this commit don't have strict creation order, this is OK. Perhaps we can use the VSS version you described above to encourage people to upgrade once we start relying on `PaginatedKVStore` in LDK Node.

@benthecarman let me know what you think.

tankyleo: @tnull I ping you on the response above in case it helps bubble this up in your inbox :)

benthecarman: Ah, I didn't know we could use `BIGSERIAL` for non-primary keys (in sqlite you can't). 

Yeah this might be better. On backwards compat, we still should be able to backfill the column by sorting by creation time, however, the migration that claude generated for this is pretty big/ugly.

benthecarman:
> I do however wonder if we should finally add a protocol version to VSS?

Yeah i think this makes sense probably for a follow up tho

benthecarman:
Pushed to use the monotonic counter. Just did the migration by adding the column and letting postgres backfill it in scan order. Doing it ourselves by sorting by creation time got really ugly and after talking with @tankyleo, didn't seem worth it.

tankyleo (commented):
LGTM will take another pass tomorrow

proto/vss.proto
tnull: > Thank you tnull for raising this point. I did some back-and-forth with claude, and yea this is interesting !
> 
> My one question at this point is backwards compat. Do you have thoughts on this point ? I'm thinking if keys created before this commit don't have strict creation order, this is OK. Perhaps we can use the VSS version you described above to encourage people to upgrade once we start relying on `PaginatedKVStore` in LDK Node.
> 
> @benthecarman let me know what you think.

I think the backfill should still handle it for the most part? But yeah, apart from that I think it might be good to lean on the versioning byte for this going forward, and make sure we publish vss-server v0.1 (with protocol versioning support) prior to LDK Node v0.8?

tnull, line 213: These docs are now inaccurate, no?

rust/api/src/kv_store_tests.rs
```diff
@@ -506,6 +505,66 @@ pub trait KvStoreTestSuite {
 		Ok(())
 	}
 
+	async fn list_should_return_results_ordered_by_creation_time() -> Result<(), VssError> {
```
tnull, line 508: Same here and below, the test names seems inaccurate now that we don't sort by time?

rust/api/src/types.rs
```diff
@@ -212,14 +212,16 @@ pub struct ListKeyVersionsRequest {
 #[allow(clippy::derive_partial_eq_without_eq)]
 #[derive(Clone, PartialEq, ::prost::Message)]
 pub struct ListKeyVersionsResponse {
-	/// Fetched keys and versions.
+	/// Fetched keys and versions, ordered by creation time (newest first).
```
tnull, line 215: Also inaccurate now.

rust/impls/src/postgres_store.rs
```diff
@@ -33,6 +33,30 @@ pub(crate) struct VssDbRecord {
 const KEY_COLUMN: &str = "key";
 const VALUE_COLUMN: &str = "value";
 const VERSION_COLUMN: &str = "version";
+const SORT_ORDER_COLUMN: &str = "sort_order";
+
+const CURRENT_PAGE_TOKEN_VERSION: char = '0';
```
tnull: Hmm, I think now that we don't have extra semantics, we should be good to drop the page token versioning byte and all associated logic again and simply use the `sort_order` as page token? (Especially given that we're discussing adding a protocol-level version, which we could also use if we'd ever find that we need to switch page token semantics again?)

README.md
```diff
@@ -48,7 +48,8 @@ integration with different cloud providers. As long as the API contract is imple
 effortlessly switch between different instances of VSS.
 
 VSS ships with a PostgreSQL implementation by default and can be hosted in your favorite infrastructure/cloud provider
-(AWS/GCP) and its backend storage can be switched with some other implementation for KeyValueStore if needed.
+(AWS/GCP). The backend storage can be switched with another implementation, but it must support ordering by creation
+time, a simple key-value store is not sufficient.
```
tnull, line 52: No, we don't sort by creation time anymore. Might be good to be more accurate here, too.

tnull (commented):
Generally looks good I think, but there are some follow-up changes to we should do, now that we don't use creation time anymore.

proto/vss.proto
tankyleo: Our implementation of this contract changed, but the contract remains the same, I think this is still accurate. Same for the other comments below.

rust/impls/src/postgres_store.rs
tankyleo: Makes sense to me yes given the upcoming protocol-level version

benthecarman:
@tnull all your comments here are wrong it seems, we still do sort by creation time, we just use a counter in the db.

tnull:
> @tnull all your comments here are wrong it seems, we still do sort by creation time, we just use a counter in the db.

I guess it depends on how literal you want to take the term 'creation time'? Fine to leave the docs if you think the slight semantic difference doesn't matter, but IMO we should still drop the now-unnecessary semantics in the page token itself (https://github.com/lightningdevkit/vss-server/pull/96#discussion_r3109490289)?

benthecarman:
Okay removed version number from page token

tankyleo (commented):
I ran this through gpt-5.5 xhigh, and it found these two things, sorry I didn't do this earlier.

  - Medium: next_page_token now exposes the raw global sort_order. Since sort_order is a global BIGSERIAL UNIQUE (rust/impls/src/migrations.rs:38) and the token is just sort_order.to_string() (rust/impls/src/postgres_store.rs:40), clients can infer service-wide write volume and gaps caused by other
    tenants. For a multi-user storage service, make the token opaque or use a per-user/store ordering value.
  - Low: negative page_size is still unvalidated, and this commit makes page_size = -1 silently return an empty successful page because fetch_limit becomes 0 (rust/impls/src/postgres_store.rs:688). Other negative values still become PostgreSQL limit errors. Reject page_size < 0 with
    InvalidRequestError before computing limit.

Curious what you think of the first one here, it's a good point I find.

benthecarman:
> Curious what you think of the first one here, it's a good point I find.

I had similar thought but I don't really seem the harm. Maybe you can infer something about user activity but not really the end of the world. Doing a per user token doesn't totally work because the big serial is across the whole table, we could encrypt it or something but it is nice that the user can easily compare between 2 tokens

tankyleo (approved):
Sounds good thanks again for the PR

tnull (approved):
Thanks!

@tankyleo Will you pick up the protocol version follow-up?

PR #97, opened 2026-04-03
Add 60 min timeout to ldk-node-integration CI

benthecarman:
This job ran for 6 hours in #96, should have a timeout to prevent this


tankyleo (approved):
Thanks, I need to find out what's pushing it past this time.

PR #99, opened 2026-05-06
Add /metrics endpoint

tankyleo:
Compatible with prometheus

### DIFF
--- a/rust/api/Cargo.toml
+++ b/rust/api/Cargo.toml
@@ -12,7 +12,6 @@ rand = { version = "0.8.5", optional = true}
 
 [target.'cfg(genproto)'.build-dependencies]
 prost-build = { version = "0.11.3" }
-reqwest =  { version = "0.11.13", default-features = false, features = ["rustls-tls", "blocking"] }
 
 [dev-dependencies]
 
--- a/rust/api/build.rs
+++ b/rust/api/build.rs
@@ -1,7 +1,7 @@
 #[cfg(genproto)]
 extern crate prost_build;
 #[cfg(genproto)]
-use std::{env, fs, fs::File, path::Path};
+use std::{env, fs, path::Path};
 
 /// To generate updated proto objects, run `RUSTFLAGS="--cfg genproto" cargo build`
 fn main() {
@@ -11,10 +11,8 @@ fn main() {
 
 #[cfg(genproto)]
 fn generate_protos() {
-	download_file(
-		"https://raw.githubusercontent.com/lightningdevkit/vss-server/7f492fcac0c561b212f49ca40f7d16075822440f/app/src/main/proto/vss.proto",
-		"src/proto/vss.proto",
-	).unwrap();
+	fs::create_dir_all("src/proto").unwrap();
+	fs::copy("../../proto/vss.proto", "src/proto/vss.proto").unwrap();
 
 	prost_build::Config::new()
 		.bytes(&["."])
@@ -23,13 +21,4 @@ fn generate_protos() {
 	println!("OUT_DIR: {}", &env::var("OUT_DIR").unwrap());
 	let from_path = Path::new(&env::var("OUT_DIR").unwrap()).join("vss.rs");
 	fs::copy(from_path, "src/types.rs").unwrap();
-}
-
-#[cfg(genproto)]
-fn download_file(url: &str, save_to: &str) -> Result<(), Box<dyn std::error::Error>> {
-	let mut response = reqwest::blocking::get(url)?;
-	fs::create_dir_all(Path::new(save_to).parent().unwrap())?;
-	let mut out_file = File::create(save_to)?;
-	response.copy_to(&mut out_file)?;
-	Ok(())
 }
--- a/rust/api/src/lib.rs
+++ b/rust/api/src/lib.rs
@@ -20,7 +20,6 @@ pub mod error;
 pub mod kv_store;
 
 /// Contains request/response types generated from the API definition of VSS.
-#[rustfmt::skip]
 pub mod types;
 
 #[cfg(feature = "_test_utils")]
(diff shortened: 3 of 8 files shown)


rust/api/src/lib.rs
```diff
@@ -20,6 +20,7 @@ pub mod error;
 pub mod kv_store;
 
 /// Contains request/response types generated from the API definition of VSS.
+#[rustfmt::skip]
```
tnull: Why skip rather than formatting post-generation?

rust/api/build.rs
```diff
@@ -11,10 +11,8 @@ fn main() {
 
 #[cfg(genproto)]
 fn generate_protos() {
-	download_file(
-		"https://raw.githubusercontent.com/lightningdevkit/vss-server/7f492fcac0c561b212f49ca40f7d16075822440f/app/src/main/proto/vss.proto",
```
tnull, line 15: AFAIU, the original intention here was to pin the protocol version to a specific commit. But you're right that we probably should do that via the new protocol versioning, and proper `vss-server` tags/releases.

rust/api/Cargo.toml
```diff
@@ -12,7 +12,6 @@ rand = { version = "0.8.5", optional = true}
 
 [target.'cfg(genproto)'.build-dependencies]
 prost-build = { version = "0.11.3" }
-reqwest =  { version = "0.11.13", default-features = false, features = ["rustls-tls", "blocking"] }
```
tnull, line 15: 🎉

rust/server/src/vss_service.rs
```diff
@@ -80,6 +80,14 @@ impl Service<Request<Incoming>> for VssService {
 			let prefix_stripped_path = path.strip_prefix(BASE_PATH_PREFIX).unwrap_or_default();
 
 			match prefix_stripped_path {
+				"/version" => {
```
tnull: Adding this endpoint is great, but shouldn't we also add a (I guess optional, for backwards compat) `protocol_version` field to all other messages? Otherwise, if we require the client to always manually first call this endpoint and wait on the response a) it's error prone b) it would require another full RTT on startup which is not great. So maybe we make version negotiation implicit and only fail the flow on mismatch? Thoughts?

rust/api/src/lib.rs
tankyleo: So that we don't need to run format post-generation ? Happy to drop if you prefer.

rust/server/src/vss_service.rs
tankyleo: See below, I add a `vss-protocol-version` header to every response, clients can read the protocol version from any response they get back from the server.

rust/api/src/lib.rs
tnull: Yeah, it's always a debate whether to format or not format auto-generated code. I always lean towards yes, others towards no. 

In the end it doesn't matter much, so I'd say I'd leave it up to you / your preference. Feel free to leave as-is.

rust/server/src/vss_service.rs
tnull: Cool. Now that we did this, maybe the endpoint could be a more general `/health` or `/status` endpoint? Or maybe this shouldn't even be a VSS-style proto response but rather a prometheus-compatible endpoint then?

tankyleo: See below, I remove `/version`, and add `/metrics`, which offers a prometheus-style gauge for now

rust/api/src/lib.rs
tankyleo: See below, I removed the skip and now format this file too.

rust/server/src/vss_service.rs
```diff
@@ -218,29 +228,35 @@ async fn handle_request<
...
 	}
 }
 
+fn build_response(status_code: StatusCode, body: Bytes) -> Response<Full<Bytes>> {
+	Response::builder()
+		.status(status_code)
+		.header(PROTOCOL_VERSION_HEADER, PROTOCOL_VERSION.as_bytes())
+		.body(Full::new(body))
+		// unwrap safety: body only errors when previous chained calls failed.
+		.unwrap()
```
tnull, line 257: Since we're touching this, should we avoid the `unwrap` and start bubbling up the error?

rust/server/src/vss_service.rs
```diff
@@ -23,6 +23,8 @@ use log::{debug, trace};
 use crate::util::KeyValueVecKeyPrinter;
 
 const MAXIMUM_REQUEST_BODY_SIZE: usize = 1024 * 1024 * 1024;
+const PROTOCOL_VERSION_HEADER: &str = "vss-protocol-version";
+const PROTOCOL_VERSION: &str = "0.1.0";
```
tnull: Hmm, protocol version usally are simple integer numbers, given they are only changed if we break the API. Do we think SemVer makes sense here?

tnull (commented):
Looks good, feel free to squash.

rust/server/src/vss_service.rs
tankyleo: Done below thank you

rust/server/src/vss_service.rs
tankyleo: Hmm so far, I buy the unwrap safety do you not buy it ? I carefully checked all the previous calls, and I don't see any panics possible here.

tankyleo:
Thanks for the review tnull, I appended commit so that the gauge metric now tracks a successful roundtrip to the `KvStore` backend let me know what you think.

tankyleo:
I'm still debating that last commit, as the metrics endpoint is unauthenticated, so with JWTs someone could hit the database for free.

tankyleo:
Ended up dropping it, now only changed to `const PROTOCOL_VERSION: &str = "0";` from last review.

tnull (approved):
LGTM, I assume versioning support in `vss-client` is up next?

tankyleo:
> LGTM, I assume versioning support in `vss-client` is up next?

Yes thanks for the review

PR #100, opened 2026-05-25
Add `vss_client_v050_compatibility` test

tankyleo:
```
    Add `vss_client_v050_compatibility` test

    We recently added the requirement that VSS servers return key-version
    lists in creation order. We also added a version header to all VSS
    responses such that clients requiring creation-order lists can ensure
    the VSS server they talk to provides this guarantee.

    These VSS server changes remain fully backwards compatible with
    previously-released VSS clients. Here we add a test that ensures that
    this is true against the last VSS client release made prior to these
    changes.

    This new test also makes sure that any future changes to VSS server
    remain backwards compatible with VSS client v0.5.0.

    AI wrote the test.
```

### DIFF
--- a/rust/server/tests/vss_client_compatibility.rs
+++ b/rust/server/tests/vss_client_compatibility.rs
@@ -1,21 +1,25 @@
-//! Compatibility shakedown for the pinned vss-client-ng v0.5.0 dependency against current
-//! vss-server master. This test assumes a no-auth VSS server is already running at
-//! `localhost:8080` and exercises a full client lifecycle through the public v0.5.0 client API:
-//! empty listing, missing-key reads, conditional and non-conditional writes, gets, conflict
-//! handling, transactional put/delete, direct deletes, paginated listing, and cleanup.
+//! Compatibility shakedown for vss-client-ng dependencies against current vss-server master. This
+//! test assumes a no-auth VSS server is already running at `localhost:8080` and exercises a full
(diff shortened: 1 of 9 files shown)


rust/server/tests/vss_client_compatibility.rs
```diff
@@ -0,0 +1,326 @@
...
+const KEY_EPSILON: &str = "compat/epsilon";
+const KEY_GAMMA: &str = "compat/gamma";
+const KEY_OUTSIDE_PREFIX: &str = "outside-prefix";
+const KEY_STALE_GLOBAL: &str = "compat/stale-global";
+const KEY_THETA: &str = "compat/theta";
+const KEY_PREFIX: &str = "compat/";
+const GLOBAL_VERSION_KEY: &str = "global_version";
+const LIST_PAGE_SIZE: i32 = 2;
+
+#[tokio::test]
```
tnull, line 37: > - [P2] Gate the live-server compatibility test — /home/tnull/workspace/vss-server/rust/server/tests/vss_client_v050_compatibility.rs:31-32
>     When cargo test -p vss-server is run outside the new workflow, no server is launched, but this test still posts to the hard-coded localhost:8080 endpoint
>     and fails with Connection refused. That regresses the normal package test command; please gate/ignore this live-server test and opt into it from server-
>     tests, or spawn the server inside the test.

tnull (commented):
one nit, otherwise LGTM

tnull (approved):

PR #102, opened 2026-06-04
Cut 0.1.0-alpha.0

tankyleo:


Cargo.toml
```diff
@@ -4,6 +4,7 @@ members = ["server", "api", "impls", "auth-impls"]
 default-members = ["server"]
 
 [workspace.package]
+version = "0.1.0-alpha.1"
```
tnull: nit: `-alpha.1` is usually the second alpha release, as `-alpha` or `alpha.0` is usually the first.

CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
+# 0.1.0-alpha.1 - Jun. 4, 2026
+
+Initial alpha release of the Versioned Storage Service server.
+
+This release is source-only; no crates.io packages, prebuilt binaries, or Docker images are published.
```
tnull: This doesn't belong in a CHANGELOG, especially since package publication would happen *after* this is merged anyways (and we *should* publish to crates.io, no?)

CHANGELOG.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Added
+
+- VSS service implementing VSS protocol version 0.
```
tnull: I would keep this more succinct as it's just "what happened so far". For changelog-style entries we'd otherwise want to highling specific features/bugfixes and link to specific PRs for further reference.

tnull (commented):
Two minor comments on the changelog, and starting with `-alpha.1` is a bit uncommon, otherwise looks good.

CHANGELOG.md
tankyleo: If we were to publish `vss-server` to crates.io, as far as I see this would also require us to publish the path-dependencies in this workspace, `api`, `auth-impls`, and `impls`.

Two options from here: 1) publish the path-dependencies to crates.io, 2) fold the path dependencies into a single crate.

I was not excited by either of them yesterday, but we can discuss further let me know what you think.

Cargo.toml
tankyleo: Sounds good I'll use alpha.0

CHANGELOG.md
tnull: What's the issue with 1) (we might need to rename the crates to something unique, but other than that)?  

2) would also be an option, IMO, but we could still consider moving `vss-client` into this repo and then it could maybe reuse `api` or other crates in its tests if we keep it separate?

Generally I think we should publish to crates.io latest with 0.1 final.

CHANGELOG.md
tankyleo: > I would keep this more succinct as it's just "what happened so far"

Can you give an example of the wording you have in mind ? Thank you sounds to me the exact line above is indeed "what's happened so far" ?

>  For changelog-style entries we'd otherwise want to highling specific features/bugfixes and link to specific PRs for further reference.

Happy to add the PR numbers yes. As for features/bug fixes isn't that what we've got below currently ? Or do you rather have in mind a brief description on a newline for each PR up to this point ?

CHANGELOG.md
tankyleo: > What's the issue with 1) (we might need to rename the crates to something unique, but other than that)?

I think I'd prefer this option to highlight the modularity of VSS-server. 

> Generally I think we should publish to crates.io latest with 0.1 final.

Ah do you mean we do not publish for this initial alpha release ?

tnull: > I think I'd prefer this option to highlight the modularity of VSS-server.

Sounds good


> Ah do you mean we do not publish for this initial alpha release ?

No, I think we should publish them, crates.io will mark them as pre-releases even. I just was under the impression that you were hesitant to publish at all, so just stating that IMO at least for the final we should.

CHANGELOG.md
tnull: Hmm, so, I think it's fine to leave as-is, though note that this will mean that the  changelog for 0.1 will be ~entirely empty. Probably fine though.

FWIW, one thing to add could be a list of contributors/authors and/or diff stats, but these are really optional.

api/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "api"
```
tnull: These crates should probably be renamed then to prepare for publishing.

tnull (commented):
Looks good, only the crates would need renaming now it seems.

CHANGELOG.md
tankyleo: > FWIW, one thing to add could be a list of contributors/authors and/or diff stats, but these are really optional.

Done below thank you

api/Cargo.toml
tankyleo: Done below, let me know if I can squash

tnull (commented):
Please squash

tankyleo:
Squashed, with an update to the release date in the changelog.

All CI workflows passed locally.

tnull (approved):

PR #114, opened 2026-09-02
pg: fix list_key_versions interpreting keys as patterns

phlip9:
Using `key LIKE $3` unescaped tells Postgres to interpret the requested key prefix as a pattern, which means %, _, and \ are interpreted as wildcards and character escapes in the original prefix rather the actual intended literal prefix.

This means a `list_key_versions(key="foo_bar", ..)` actually includes `foo1bar`, which definitely does not seem intentional.

A better approach is to just use `starts_with(key, $3)`, which the query planner understands better and doesn't require any escaping.


phlip9:
cc @tnull (can't add reviewers in the sidebar 😢)

impls/src/postgres_store.rs
```diff
@@ -691,26 +691,24 @@ where
 
 		let conn = self.pool.get().await?;
 
-		let key_like = format!("{}%", key_prefix.as_deref().unwrap_or_default());
-
 		let rows = if let Some(ref token) = page_token {
 			let page_sort_order = decode_page_token(token)?;
-			let stmt = "SELECT key, version, sort_order FROM vss_db WHERE user_token = $1 AND store_id = $2 AND sort_order < $3 AND key LIKE $4 AND key != $5 ORDER BY sort_order DESC LIMIT $6";
+			let stmt = "SELECT key, version, sort_order FROM vss_db WHERE user_token = $1 AND store_id = $2 AND sort_order < $3 AND starts_with(key, $4) AND key != $5 ORDER BY sort_order DESC LIMIT $6";
```
tankyleo: Many thanks can you also add test coverage for the diff in this branch here ?

impls/src/postgres_store.rs
```diff
@@ -691,26 +691,24 @@ where
...
 			let params: Vec<&(dyn tokio_postgres::types::ToSql + Sync)> = vec![
```
tankyleo: nit: in passing could you drop the vec here and below ?

tankyleo (commented):
thanks sorry for the delay this needs a rebase

tankyleo (approved):

PR #115, opened 2026-09-03
pg: prepare+cache query statements

phlip9:
Without this change, we would re-parse and re-plan each query, potentially multiple times per request (!). For simple queries, this appears to add ~100-200 us overhead. More for more complex queries.

Instead, maintain a small prepared statement cache per connection. We'll lazily prepare a query statement the first time we execute it, then use the prepared statement thereafter.

This change reduces latency by ~20-60% on my small benchmark suite, but is particularly impactful on batch conditional updates, where we were previously re-parsing and re-planning the `put` query for each item in the batch. This change improved throughput there by ~2.5x (~4.8k/s -> 12.3k/s).

Using const `&'static str` queries is really more a stylistic preference on my part. I think it adds a nice roadblock to prevent people from accidentally adding `let stmt = format!("..", untrusted_user_input)` -> SQL injection.

Migrations and other admin queries are still uncached, since they're usually only executed once.


phlip9:
Oops, looks like the new `tinyvec` release accidentally broke `alloc` + no-std. Waiting for https://github.com/Lokathor/tinyvec/pull/226

impls/src/postgres_store.rs
```diff
@@ -1143,7 +1260,27 @@ mod tests {
...
+		let stmt = "SELECT $1::BIGINT";
+
+		let mut transaction = conn.transaction().await.unwrap();
+		transaction.execute(stmt, &[&1_i64]).await.unwrap();
+		transaction.commit().await.unwrap();
+		assert_eq!(conn.statement_cache.statements.len(), 1);
+
+		let rows = conn.query(stmt, &[&2_i64]).await.unwrap();
+		assert_eq!(rows[0].get::<_, i64>(0), 2);
+		assert_eq!(conn.statement_cache.statements.len(), 1);
```
tankyleo: Recruited AI to help me expand the test like this to make sure postgres agrees we've reused the same prepared statement within and across different queries and transactions. Let me know what you think no need for attribution.
```rust
        #[tokio::test]
        async fn prepared_statements_should_be_reused_within_and_across_transactions() {
                async fn statement_usage(conn: &Client, stmt: &str) -> (String, u64) {
                        // Inspect the same session without preparing another statement for the inspection.
                        let messages = conn
                                .uncached_client
                                .simple_query(
                                        "SELECT name, statement, generic_plans + custom_plans AS uses
                                         FROM pg_prepared_statements",
                                )
                                .await
                                .unwrap();
                        let statements: Vec<_> = messages
                                .into_iter()
                                .filter_map(|message| match message {
                                        SimpleQueryMessage::Row(row) if row.get("statement") == Some(stmt) => Some((
                                                row.get("name").unwrap().to_owned(),
                                                row.get("uses").unwrap().parse::<u64>().unwrap(),
                                        )),
                                        _ => None,
                                })
                                .collect();
                        assert_eq!(statements.len(), 1, "expected one prepared statement: {statements:?}");
                        statements.into_iter().next().unwrap()
                }

                let mut conn = Client::connect(&POSTGRES_ENDPOINT, DEFAULT_DB, NoTls).await.unwrap();
                let stmt = "SELECT $1::BIGINT";

                let mut transaction = conn.transaction().await.unwrap();
                transaction.execute(stmt, &[&1_i64]).await.unwrap();
                transaction.execute(stmt, &[&2_i64]).await.unwrap();
                transaction.commit().await.unwrap();
                let (name, uses) = statement_usage(&conn, stmt).await;
                assert_eq!(uses, 2, "statement should be reused within a transaction");

                let mut transaction = conn.transaction().await.unwrap();
                transaction.execute(stmt, &[&3_i64]).await.unwrap();
                transaction.commit().await.unwrap();
                assert_eq!(statement_usage(&conn, stmt).await, (name.clone(), 3));

                let rows = conn.query(stmt, &[&4_i64]).await.unwrap();
                assert_eq!(rows[0].get::<_, i64>(0), 4);
                assert_eq!(statement_usage(&conn, stmt).await, (name.clone(), 4));

                let row = conn.query_opt(stmt, &[&5_i64]).await.unwrap().unwrap();
                assert_eq!(row.get::<_, i64>(0), 5);
                assert_eq!(statement_usage(&conn, stmt).await, (name, 5));
        }
```

impls/src/postgres_store.rs
```diff
@@ -390,13 +392,17 @@ where
 	async fn execute_non_conditional_upsert(
 		&self, transaction: &Transaction<'_>, vss_record: &VssDbRecord,
 	) -> io::Result<u64> {
-		let stmt = format!("INSERT INTO vss_db (user_token, store_id, key, value, version, created_at, last_updated_at)
-                    VALUES ($1, $2, $3, $4, {}, $5, $6)
-                    ON CONFLICT (user_token, store_id, key) DO UPDATE
-                    SET value = EXCLUDED.value, version = {}, last_updated_at = EXCLUDED.last_updated_at", INITIAL_RECORD_VERSION, INITIAL_RECORD_VERSION);
+		#[rustfmt::skip]
+		const STMT: &str = const_concat_str!(
```
tankyleo: Can we drop `const_concat_str` and instead  bind the `INITIAL_RECORD_VERSION` as a SQL parameter ? Something like this for `execute_non_conditional_upsert`, similar for `execute_conditional_insert`.

```rust
                const STMT: &str = "INSERT INTO vss_db (user_token, store_id, key, value, version, created_at, last_updated_at)
                        VALUES ($1, $2, $3, $4, $5, $6, $7)
                        ON CONFLICT (user_token, store_id, key) DO UPDATE
                        SET value = EXCLUDED.value, version = $5, last_updated_at = EXCLUDED.last_updated_at";
                let num_rows = transaction
                        .execute(
                                STMT,
                                &[
                                        &vss_record.user_token,
                                        &vss_record.store_id,
                                        &vss_record.key,
                                        &vss_record.value,
                                        &i64::from(INITIAL_RECORD_VERSION),
                                        &vss_record.created_at,
                                        &vss_record.last_updated_at,
                                ],
                        )
                        .await
                        .map_err(|e| {
                                Error::new(ErrorKind::Other, format!("Database operation failed. {}", e))
                        })?;
                Ok(num_rows)
```

impls/src/postgres_store.rs
```diff
@@ -752,10 +752,14 @@ mod tests {
...
 	use bytes::Bytes;
+	use std::sync::LazyLock;
 	use tokio::sync::OnceCell;
 	use tokio_postgres::NoTls;
 
-	const POSTGRES_ENDPOINT: &str = "postgresql://postgres:postgres@localhost:5432";
+	static POSTGRES_ENDPOINT: LazyLock<String> = LazyLock::new(|| {
+		std::env::var("POSTGRES_ENDPOINT")
+			.unwrap_or_else(|_| "postgresql://postgres:postgres@localhost:5432".to_string())
+	});
```
tankyleo: nit: can we drop the `LazyLock` and the associated borrows below and use this?
```rust
        const POSTGRES_ENDPOINT: &str = match option_env!("POSTGRES_ENDPOINT") {
                Some(endpoint) => endpoint,
                None => "postgresql://postgres:postgres@localhost:5432",
        };
```

This shifts it to a compile-time read, but seems this fits your usecase.

tankyleo (commented):
Thanks again sorry for the delay, feel free to rebase.

PR #116, opened 2026-09-10
cargo: make vss-server/impls easier to consume as a library

phlip9:
### server+impls: make postgres native TLS optional

The `native-tls` crate adds a ton of build headache that I'd like to
avoid if I'm not actually using it. This diff makes it possible to
disable it.

### cargo: relax direct dependency version requirements

Make it easier to consume VSS crates as a library. I've limited direct
crate dep semver versions to the first MSRV+semver compatible crate that
compiles and passes all tests.

* tokio-v1.30 is the first version that supports `OnceCell::const_new`
  without the `parking_lot` crate enabled


phlip9:
Looks like vss-server also needs something like https://github.com/lightningdevkit/ldk-node/pull/1092 to fix MSRV CI checks

tnull (commented):
Not opposed to the change, but could expand a bit on a) what headache you're referring to exactly and b) what's your use case for using the crates independently from the `vss-server` binary?

Changes themselves LGTM I think.

impls/src/postgres_store.rs
```diff
@@ -19,8 +17,6 @@ use tokio_postgres::{error, Client, NoTls, Socket, Transaction};
 
 use log::{debug, info, warn};
 
-pub use native_tls::Certificate;
```
tnull, line 22: I think it would be good to keep this re-export (even if only behind the feature).

phlip9:
Sure, let me expand on this a bit:

a) We try really hard to use only one TLS stack (rustls) as much as we can. This is for hardening (avoid exposing more non-memory safe code to the open internet), build convenience (it's slightly annoying to configure openssl+pkgconfig in nix builds and dev machines), dependency minimization, reduced binary bloat, and reduced CI build time. So pulling in `openssl` and the `native-tls` crate is something I would like to avoid.

b) We're now running an internal VSS (inside our VPC with mTLS auth). It was actually really nice just pulling in the `VssService` hyper service from `server/src/vss_service.rs` and serving it on top of all our existing Rust tooling. This integrates cleanly into our e2e tests, gets all our logs+traces+metrics tooling, mTLS config, instrumented allocator, etc ~ for free. It's not that we can't do this as a separate "opaque" service behind nginx, but then it's definitely harder to test.

tnull:
> Sure, let me expand on this a bit:
> 
> a) We try really hard to use only one TLS stack (rustls) as much as we can. This is for hardening (avoid exposing more non-memory safe code to the open internet), build convenience (it's slightly annoying to configure openssl+pkgconfig in nix builds and dev machines), dependency minimization, reduced binary bloat, and reduced CI build time. So pulling in `openssl` and the `native-tls` crate is something I would like to avoid.
> 
> b) We're now running an internal VSS (inside our VPC with mTLS auth). It was actually really nice just pulling in the `VssService` hyper service from `server/src/vss_service.rs` and serving it on top of all our existing Rust tooling. This integrates cleanly into our e2e tests, gets all our logs+traces+metrics tooling, mTLS config, instrumented allocator, etc ~ for free. It's not that we can't do this as a separate "opaque" service behind nginx, but then it's definitely harder to test.

Okay, makes sense (though I'm not sure where I stand on the `rustls` vs `openssl` arguments, but that is for an orthogonal discussion, trade-offs everywhere).

So, happy to have this land from my side, but @tankyleo still needs to review.

phlip9:
Pushed a separate PR to fix MSRV CI issues: https://github.com/lightningdevkit/vss-server/pull/117

impls/Cargo.toml
```diff
@@ -9,17 +9,21 @@ license.workspace = true
...
-native-tls = { version = "0.2.14", default-features = false }
-postgres-native-tls = { version = "0.5.2", default-features = false, features = ["runtime"] }
-log = { version = "0.4.29", default-features = false }
+chrono = "0.4.16"
+tokio-postgres = { version = "0.7.15", features = ["with-chrono-0_4"] }
+bytes = "1"
+tokio = { version = "1.30", default-features = false, features = ["rt", "macros"] }
+native-tls = { version = "0.2.4", optional = true, default-features = false }
+postgres-native-tls = { version = "0.5", optional = true, default-features = false, features = ["runtime"] }
+log = { version = "0.4.8", default-features = false }
```
tankyleo: Looks like version 0.4.6 still works here

server/Cargo.toml
```diff
@@ -21,15 +22,15 @@ impls = { workspace = true }
...
-log = { version = "0.4.29", default-features = false, features = ["std"] }
-chrono = { version = "0.4", default-features = false, features = ["clock"] }
-rand = { version = "0.9.2", default-features = false }
+hyper-util = { version = "0.1", default-features = false, features = ["tokio"] }
+tokio = { version = "1.30", default-features = false, features = ["time", "signal", "rt-multi-thread", "macros"] }
+prost = { version = "0.11", default-features = false, features = ["std"] }
+bytes = "1"
+serde = { version = "1.0.184", default-features = false, features = ["derive"] }
+toml = { version = "0.8", default-features = false, features = ["parse"] }
+log = { version = "0.4.8", default-features = false, features = ["std"] }
```
tankyleo: Similar here, 0.4.6 works

impls/src/lib.rs
```diff
@@ -5,6 +5,8 @@
 //! simplify the development process for Lightning wallets by providing a secure means to store
 //! and manage the essential state required for Lightning Network (LN) operations.
 //!
+//! The `postgres-native-tls` feature enables the native TLS backend and is enabled by default.
```
tankyleo: nit: this seems a little out of place, can we delete it ?

impls/Cargo.toml
```diff
@@ -9,15 +9,19 @@ license.workspace = true
...
 async-trait = "0.1.13"
 api = { workspace = true }
 chrono = "0.4.16"
 tokio-postgres = { version = "0.7.15", features = ["with-chrono-0_4"] }
 bytes = "1"
 tokio = { version = "1.30", default-features = false, features = ["rt", "macros"] }
-native-tls = { version = "0.2.4", default-features = false }
-postgres-native-tls = { version = "0.5", default-features = false, features = ["runtime"] }
+native-tls = { version = "0.2.4", optional = true, default-features = false }
+postgres-native-tls = { version = "0.5", optional = true, default-features = false, features = ["runtime"] }
```
tankyleo: Can we correct the prerequisites section in `docs/getting-started.md` to clarify that OpenSSL development/runtime libraries are now only required for JWT authentication or PostgreSQL TLS support ?

Something like:

```
## Prerequisites

- Rust and Cargo, using at least the repository MSRV of 1.85.0.
- PostgreSQL 15 or newer.
- OpenSSL development/runtime libraries for JWT authentication or PostgreSQL TLS support, both included in default builds.
```

server/Cargo.toml
tankyleo: We don't update the `Cargo.lock` file in commit 42388338c0588c6c5987c81171579506300fab5c so I am worried we may ship a piece of code in the future that no longer works with the declared minimum versions in this commit.

Is it worth adding some automated testing for this case ?

tankyleo (commented):
[Some buggy github stuff ignore this]

impls/src/lib.rs
phlip9: yeah now that I look at it, it definitely feels a bit tacked on. I'll remove this line.

~~on a related note, I notice there's no docs for the `jwt` or `sigs` features (now also none for `postgres-native-tls`).~~ EDIT: nvm, I see brief mention in getting-started.

when looking at a new crate, I usually skim through the docs for an "Optional features" list like <https://docs.rs/hyper/latest/hyper/#optional-features>. so we should probably add similar somewhere. normally I would put that in the README.md or top-level crate `lib.rs`.

though, the current README.md reads more like a design doc than a project landing page. it might be worth placing more of the content from doc/getting-started.md more upfront in the README and put more of the current README into a doc/design.md

phlip9: added a small section to getting-started

server/Cargo.toml
phlip9: sure, I usually do something like this in CI:

```bash
# Remove dev-deps from Cargo.toml to prevent `cargo update` determining minimal
# versions based on dev-deps.
cargo hack --remove-dev-deps --workspace

# Resolve direct dependencies using the min. version in our Cargo.toml's.
RUSTC_BOOTSTRAP=1 cargo update -Z direct-minimal-versions

cargo check --workspace
```

I'll add this

phlip9: done

PR #117, opened 2026-09-22
ci: fix MSRV tests. make more reliable.

phlip9:
This PR should make MSRV tests in CI more reliable, with less package pinning required.

First, migrate the workspace to edition 2024 and resolver-v3, so we enable the MSRV-aware resolver by default:
<https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions>

This is stable since 1.84, so shouldn't require an MSRV bump.

Then manually enable the MSRV-aware resolver for `ldk-node` build in the integration test. This should reduce the number of pinned packages required to get `ldk-node` to build cleanly.

### Alternative

Add `cargo update -p encoding-rs --precise "0.8.35" --verbose` MSRV pin to all CI workflows


tnull (approved):

tankyleo (approved):
thank you !

